PR #16, opened 2024-03-28
Update getEdgesDirections queries

Cole-Greer:
*Issue #, if available:*

*Description of changes:*

Replaces the queries to fetch edge directions while building the schema. The old queries worked by testing for the existance of any possible edge, which resulted in a large number of queries for graphs with lots of node and edge labels.

The new query instead iterates through all the edge labels, and runs a query to fetch the corresponding from-label to-label pairs.

### DIFF
--- a/src/NeptuneSchema.js
+++ b/src/NeptuneSchema.js
@@ -132,28 +132,22 @@ async function getEdgesNames() {
 }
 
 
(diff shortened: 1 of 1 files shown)


andreachild:
VOTE +1 with the assumption that parameterized queries will be done as part of a different PR

PR #27, opened 2024-10-11
Change integration tests to use injected environment variables

andreachild:
Removed duplicates of airports.source.schema.graphql and changed references to use a common file. Changed integration test references to neptune db host and port to have values injected from environment variables so that contributors do not have to modify the case files to execute tests, which could lead to accidental commit of real URLs.

Fixed tests failing due to missing comma in json file, typo in referenced file suffix, and incorrect order of test cases (delete, create changed to create, delete).


test/testLib.js
```diff
@@ -20,7 +20,10 @@ async function queryNeptune(q, language, host, port, param) {
 
 
 function readJSONFile(fileName) {
-    return JSON.parse(fs.readFileSync(fileName, 'utf8'));
+    let text = fs.readFileSync(fileName, 'utf8');
+    text = text.replace('<AIR_ROUTES_DB_HOST>', process.env.AIR_ROUTES_DB_HOST);
+    text = text.replace('<AIR_ROUTES_DB_PORT>', process.env.AIR_ROUTES_DB_PORT);
```
kmcginnes, line 25: Not necessary, but it would be nice to have a good error message if these environment values are not defined. From what I see here it would just put `undefined` in place of the tokens and fail at some later point.

kmcginnes (approved):
I had one suggestion, but it is not necessary

PR #29, opened 2024-10-11
Change neptune queries to use parameters

andreachild:
Changed neptune queries to use parameters when able to prevent open cypher query injection. Added placeholder function for sanitization of query text that cannot be parameterized (such as node and edge labels in match clause). Sanitization logic is still to be determined after consultation with AWS.


src/NeptuneSchema.js
```diff
@@ -57,34 +57,50 @@ function consoleOut(text) {
...
     } else {
-        try {        
-        const response = await axios.post(`https://${HOST}:${PORT}/${language}`, `query=${encodeURIComponent(q)}`);
-        return response.data;    
+        try {
+            let data = {
+                query: query,
+                parameters: params
+            };
+            const response = await axios.post(`https://${HOST}:${PORT}/${language}`, data);
```
kmcginnes, line 82: I'm curious if the change to the data parameter results in a change to how the request body is serialized. I couldn't find the default serialization method for axios. It might serialize the JSON object in to URL form encoding. But it might keep it JSON, which would be a change.

You can pass the serialization method if you want to be explicit.

https://axios-http.com/docs/post_example

src/NeptuneSchema.js
```diff
@@ -196,9 +212,10 @@ function addUpdateEdgeProperty(edgeName, name, value) {
 
 
 async function getEdgeProperties(edge) {
-    let query = `MATCH ()-[n:${edge.label}]->() RETURN properties(n) as properties LIMIT ${SAMPLE}`;        
+    let query = `MATCH ()-[n:${sanitize(edge.label)}]->() RETURN properties(n) as properties LIMIT $sample`;
+    let parameters = `{"sample": ${SAMPLE}}`;
```
kmcginnes, line 216: I'm unfamiliar with the parameters format, but it looks a lot like JSON. Would it make more sense to construct a proper JSON object, the stringify it right before sending in the request? That would ensure the syntax is correct.

kmcginnes (commented):
Left a couple questions and suggestions.

src/NeptuneSchema.js
andreachild: The `aws4Interceptor` complained when I set it to form so I'm guessing it actually changed it to json under the hood. I set it to json instead to be explicit.

src/NeptuneSchema.js
andreachild: Changed to instantiate as object and then use JSON.stringify as suggested.

kmcginnes: What you have works, but I was thinking you would move the `JSON.stringify()` in to `function queryNeptune()`. You could even support both a string parameter and an object.

Something like this:

```js
async function queryNeptune(query, params) {
    // Ensure params is a string value
    const paramsString = typeof params === 'string'
        ? params
        : JSON.stringify(params)
    
    if (useSDK) {
        const response = await queryNeptuneSDK(query, paramsString);
        return response;
    } else {
        try {
            let data = {
                query: query,
                parameters: paramsString
            };
            const response = await axios.post(`https://${HOST}:${PORT}/${language}`, data, {
                headers: {
                    'Content-Type': 'application/json'
                }
            });
        return response.data;
        } catch (error) {
            console.error("Http query request failed: ", error.message);
            consoleOut("Trying with the AWS SDK");
            const response = await queryNeptuneSDK(query, paramsString);
            useSDK = true;
            return response;
        }
    } 
}
```

andreachild: Thanks for pointing that out, I should have noticed that delaying JSON.stringify is better happening later than sooner .

I changed it to something similar except to set the default params value to empty object instead of supporting both object and string params as I didn't see the need to support both types at this point in time. Callers should always be passing an object that is convertible to JSON string as you pointed out in your first comment so this makes more sense to me.

kmcginnes (approved):

PR #30, opened 2024-10-11
Fix for typos

andreachild:
Fix for typos mostly in documentation but also a couple command line arguments and one model-scoped variable that had 'neptume' instead of 'neptune'.


README.md
```diff
@@ -77,7 +77,7 @@ To rollback, removing all the AWS resources run:
 `neptune-for-graphql --remove-aws-pipeline-name` <*your-new-GraphQL-API-name*>
 
 #### References:
-- [here](/doc/routesExample.md) an example using the Air Routes data on Amazon Neptune, showing the outputs of the utility.
+- [here](/doc/routesExample.md) is an example using the Air Routes data on Amazon Neptune, showing the outputs of the utility.
```
kmcginnes, line 80: Should "here" be capitalized?

README.md
```diff
@@ -89,7 +89,7 @@ You can start from an empty Neptune database and use a GraphQL API to create the
 `neptune-for-graphql --input-schema-file `<*your-graphql-schema-file*>` --create-update-aws-pipeline --create-update-aws-pipeline-name` <*your-new-GraphQL-API-name*>` --create-update-aws-pipeline-neptune-endpoint` <*your-neptune-database-endpoint:port*>  ` --output-resolver-query-https`
 
 #### References:
-- [here](https://github.com/aws/amazon-neptune-for-graphql/blob/main/doc/todoExample.md) an example using a TODO GraphQL schema, showing the outputs of the utility.
+- [here](https://github.com/aws/amazon-neptune-for-graphql/blob/main/doc/todoExample.md) is an example using a TODO GraphQL schema, showing the outputs of the utility.
```
kmcginnes, line 92: Should "here" be capitalized?

README.md
```diff
@@ -160,7 +160,7 @@ type Mutation {
 }
 ```
 
-You can add query or mututation using a Gremlin query. At this time Gremlin queries are limited to return *scalar* values, or *elementMap()* for a single node, or *elementMap().fold()* for a list of nodes.
+You can add a query or mutation using a Gremlin query. At this time Gremlin queries are limited to return *scalar* values, or *elementMap()* for a single node, or *elementMap().fold()* for a list of nodes.
```
kmcginnes, line 163: This might be out of scope for your changes, but the code could be formatted with back ticks instead of italics.

> ...return `scalar` values, or `elementMap()` for a single node, or `elementMap().fold()` for a list of nodes.

kmcginnes (commented):
These changes are great. I offered a few additional suggestions.

Also worth considering is that the fix for the typo in the command line argument is technically a breaking change. I assume we are good with that, but I wanted to ask the question.

andreachild:
> These changes are great. I offered a few additional suggestions.
> 
> Also worth considering is that the fix for the typo in the command line argument is technically a breaking change. I assume we are good with that, but I wanted to ask the question.

I have brought it up before with stakeholders that I intended to fix the command line option spelling and there didn't seem to be concerns (other than to ensure the relevant documentation is also updated) but I will call out it's a breaking change as well and see what the feedback is.

README.md
andreachild: Changed to capital

README.md
andreachild: Changed to capital

README.md
andreachild: I was about to make this change but then realized italics are used in many places of this README so would prefer not to as part of this PR. Maybe in another separate PR.

PR #34, opened 2024-10-17
Fix CDK deployment

andreachild:
Fixed CDK lambda function bugs which were causing timeout and unauthorized errors when using the deployed GraphQL API app:
- VPC config now has the appropriate subnets and security group configured
- policy actions now reference the correct aws service depending on the neptune type
- environment now references the correct region when the neptune type is neptune-graph
- refactored endpoint parsing to use new util function so that the logic is contained in place and parsing is done in one function call instead of sprinkled throughout the code
- added unit tests for endpoint parsing
- added integration test verification for CDK deployment to check that the generated CDK js file contains some expected content


kmcginnes (approved):
LGTM

PR #56, opened 2025-01-08
Save the neptune schema to file early so that it can be used for troubleshooting

andreachild:
Save the neptune schema to file early (as soon as it's available) so that it can be used for troubleshooting if errors occur later in the pipeline.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -264,6 +264,36 @@ function processArgs() {
 
 }
 
+/**
+ * Saves the neptune schema to file
+ */
+function saveNeptuneSchema() {
+    // Output Neptune schema
+    if (inputGraphDBSchema !== '') {
+        if (outputNeptuneSchemaFile === '') {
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -264,6 +264,36 @@ function processArgs() {
...
 }
 
+/**
+ * Saves the neptune schema to file
+ */
+function saveNeptuneSchema() {
+    // Output Neptune schema
+    if (inputGraphDBSchema !== '') {
+        if (outputNeptuneSchemaFile === '') {
+            if (createUpdatePipelineName === '') {
```
Cole-Greer, line 274: Is there any possibility for these variables to be undefined instead of an empty string?

andreachild: They are initialized to empty string at the start of the file

Cole-Greer (approved):

PR #57, opened 2025-01-17
Alias edges with same label as a node

andreachild:
Detect if there is an edge with the same label as a node and if so, create an alias in the graphQL schema by prefixing the edge label with an underscore. Without adding this alias the request to AWS to publish the graphQL API will fail as the schema will be invalid due to the duplication of types.


kmcginnes (approved):
LGTM

Issue #20, opened 2024-07-02
Bug: Http query request failed


danielmartincraig:
I have encountered an issue; I run neptune-for-graphql and I see that after finding all my nodes and edges via the Neptune Summary API, and after discovering the properties of the edges and nodes, I see error messages like the following begin to flow:

`Http query request failed:  Client network socket disconnected before secure TLS connection was established
Trying with the AWS SDK`

I am running the tool like so:

`neptune-for-graphql --input-graphdb-schema-neptune-endpoint xxxx-xxx-xx.us-east-1.neptune.amazonaws.com:xxxx --output-aws-pipeline-cdk-neptune-IAM`

I am using Neptune engine 1.2.1.0, with sigV4 enabled.  My cluster instance is of type 'provisioned', it's of size db.r5.4xlarge.

kenhuuu:
Just to clarify, are you saying that the utility isn't working after finding the nodes/edges and you are seeing those error messages, or are you saying that the utility does work but you are asking if you should be concerned about those error messages?

danielmartincraig:
The utility is not working; it starts logging the error and it never finishes running and never puts anything in the output/ folder.

danielmartincraig:
It just starts logging the error endlessly

PR #58, opened 2025-01-23
Cap concurrent requests to get Neptune schema

danielfinke:
*Issue #, if available:* #20

*Description of changes:*

Previously, for a graph with `n` nodes or edges, there would be `n` concurrent requests made to Neptune via `queryNeptune`. Depending on your instance class or VPC settings, you could hit errors such as `MemoryLimitExceededException` or connection limits/sockets abruptly closing, with increasing likelihood as the size of the graph increased. Initially observed abrupt socket closing just like aws/amazon-neptune-for-graphql#20.

Now, the concurrent request Promises are resolved in batches using `mapAll`. For now, the batch size is hardcoded to 20. For very small instance classes (e.g. from the "Development and testing" template), `MemoryLimitExceededException`s are still likely.

Further improvements could include:
- relating the batch size to the instance class
- allowing the user to specify the batch size in the process args
- exponential/dynamic backoff on transient errors as indicated in https://docs.aws.amazon.com/neptune/latest/userguide/errors-engine-codes.html#errors-query

### DIFF
--- a/src/test/util-promise.test.js
+++ b/src/test/util-promise.test.js
@@ -10,11 +10,12 @@ describe('mapAll', () => {
     test('should map each element in batches', async () => {
         const count = 100;
         const batchSize = 50;
+        const batchCounter = makeBatchCounter(count);
 
-        await expect(mapAll(Array(count + 1).fill(0), batchCounter(count), batchSize))
+        await expect(mapAll(Array(count + 1).fill(0), (i) => batchCounter.fn(i), batchSize))
             .rejects
             .toThrow(Error);
-        expect(count).toEqual(count);
+        expect(batchCounter.count).toEqual(count);
     });
 
     test('should reject with the same error as a rejected mapped Promise', async () => {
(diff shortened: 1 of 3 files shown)


src/util-promise.js
andreachild, line 1: Hi this looks good, thanks for finding and fixing this issue. I would only suggest to add a small unit test for basic happy path and error path scenario coverage.

Suggestion:

```
import {mapAll} from '../util-promise.js';

test('should execute callback for each element in batches', () => {
    return mapAll([1,2,3,4,5,6,7,8], doubleIt, 2).then(results => {
        expect(results).toStrictEqual([2,4,6,8,10,12,14,16]);
    })
});

test('should throw an error if one callback fails', () => {
    expect(async () => {
        await mapAll([2,4,6,3,8,10], errorIfOdd, 2);
    }).rejects.toThrow('Odd number!');
});

function doubleIt(i) {
    return i + i;
}

function errorIfOdd(i) {
    if (i % 2 !== 0) {
        throw Error('Odd number!')
    }
    return i;
}
```

src/test/util-promise.test.js
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    test('should map each element in batches', async () => {
+        const count = 100;
+        const batchSize = 50;
+
+        await expect(mapAll(Array(count + 1).fill(0), batchCounter(count), batchSize))
+            .rejects
+            .toThrow(Error);
+        expect(count).toEqual(count);
```
andreachild, line 17: This test file is looking much better than the quick suggestion that I had made. I'm a bit puzzled by this particular unit test however. It seems this line in particular is testing that the const count is equal to itself?

danielfinke: oops, broke this expectation when refactoring out `batchCounter`, pushing a fix

andreachild:
LGTM

Cole-Greer (approved):
Thanks for the contribution @danielfinke. LGTM

Issue #59, opened 2025-01-24
`@id` directive is ignored, `schemaModelValidator` always adds a new ID field


danielfinke:
When the `@id` directive is specified on a field, such as in the following type,

```graphql
type User {
  userId: ID! @id
  fullName: String
  isAdmin: Boolean
  location: String
}
```

it is ignored by `schemaModelValidator` which adds a new ID field, resulting in 2 ID fields:

`output.source.schema.graphql`
```graphql
type User {
  _id: ID! @id
  userId: ID! @id
  fullName: String
  isAdmin: Boolean
  location: String
}
```

Command: `node ./neptune-for-graphql.mjs --input-schema-file test.graphql --output-no-lambda-zip`

Expected output:

`output.source.schema.graphql`
```graphql
type User {
  userId: ID! @id
  fullName: String
  isAdmin: Boolean
  location: String
}
```

PR #60, opened 2025-01-24
Honour `@id` directive on type fields

danielfinke:
*Issue #, if available:* #59

*Description of changes:*

When the `@id` directive was specified on a field, it was ignored by `schemaModelValidator` which added a new ID field, resulting in 2 ID fields.

Now, the directive inference checks for the existence of an ID field before adding the default one.

- break out logic around ID/input fields into dedicated functions
- better inlining of input fields and values

### DIFF
--- a/src/test/schemaModelValidator.test.js
+++ b/src/test/schemaModelValidator.test.js
@@ -1,0 +1,54 @@
+import { readFileSync } from 'node:fs';
+import { loggerInit } from '../logger.js';
+import { validatedSchemaModel } from '../schemaModelValidator.js';
+import { schemaParser } from '../schemaParser.js';
+
+describe('validatedSchemaModel', () => {
+    let model;
+
+    beforeAll(() => {
+        loggerInit('./output', false, 'silent');
+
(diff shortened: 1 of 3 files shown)


kmcginnes:
Looks good to me!

+1

src/test/schemaModelValidator.test.js
```diff
@@ -0,0 +1,54 @@
...
+        const userIdFields = getIdFields(userType);
+        const groupIdFields = getIdFields(groupType);
+
+        expect(userIdFields).toHaveLength(1);
+        expect(groupIdFields).toHaveLength(1);
+        expect(userIdFields[0].name.value).toEqual('userId');
+        expect(groupIdFields[0].name.value).toEqual('_id');
+    });
+
+    test('should define the same ID fields on a type and its input type', () => {
```
andreachild, line 30: Would be nice to also add verification of the `Group` type to differentiate since it did not have a specified `id` field and it would be expected to be added to the type and input type.

andreachild:
LGTM

Cole-Greer (approved):
Thanks for the contribution @danielfinke

Issue #44, opened 2024-11-20
Query fails if filter is specified as empty


andreachild:
If a graphQL query contains a filter that is an empty object `filter: {}` , the query will fail. If the filter is removed entirely then the query is executed successfully.

![Screenshot 2024-11-20 at 12 59 23 PM](https://github.com/user-attachments/assets/ad11ef9a-b9be-4193-9e55-c82091b4799a)

![Screenshot 2024-11-20 at 12 59 47 PM](https://github.com/user-attachments/assets/370edc26-13a3-41a1-9881-09cf270c8f14)

Problem is that the resolver is resolving the graphQL query to something like this (filter value is missing opening curly brace):
```
getNodeAirports(filter:}) {
    ...
}
```

PR #61, opened 2025-01-25
Fix query from AppSync with an empty filter object

danielfinke:
*Issue #, if available:* #44

*Description of changes:*

Build the query/mutation's selection set using the `graphql` utilities, instead of building manually. This will avoid any edge cases with the provided arguments. The schema is assumed to be valid: validation should be taken care of ahead of time in `schemaModelValidator` and we also currently do not define the built-in directives (e.g. `@id`, `@relationship`) required to make it a valid schema.

- break out lookup functions for type/field definitions
- update `outputReference` resolvers
- add unit tests
- fix query result tests that were sensitive to object key order (use `toEqual` instead of strict JSON string equality)
- also check output file size of `.cjs` resolver in test `Case01.02`
- fix running test `Case01.04` (use `.cjs` extension)


andreachild:
Hi @danielfinke thanks for all of your recent contributions. This one is a bit more complex than the others so will take a bit more time to review, but I hope to review it this week.

AndreaNassisi:
Hello Daniel, thank you for submitting these changes. The team will look at them soon.

templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -20,36 +21,41 @@ const schemaDataModelJSON = `INSERT SCHEMA DATA MODEL HERE`;
...
+            event.selectionSetGraphQL !== ''
+                ? gql`${event.selectionSetGraphQL}`.definitions[0].selectionSet
+                : undefined,
+    };
+    const opSelectionSet = {
+        kind: 'SelectionSet',
+        selections: [fieldNode]
+    };
+
+    const graphQuery = resolveGraphDBQuery(print(opSelectionSet));
```
andreachild, line 58: I like the idea of building the selection set using the `graphql` utilities instead of manually concatenating strings together.  However I noticed that there is some level of inefficiency with back and forth transformations to/from `gql` object. I will try to explain my concern with example below.

This code is triggered with an app sync event like:
```
{
  field: 'getNodeAirports',
  arguments: { filter: { country: 'US' }, options: { limit: 3 } },
  selectionSetGraphQL: '{\n  code\n  city\n}',
  source: null
}
```

After some transformations, on line 44 the `fieldNode` object ends up looking like:
```

{
    "kind": "Field",
    "name": {
        "kind": "Name",
        "value": "getNodeAirports"
    },
    "arguments": [
        {
            "kind": "Argument",
            "name": {
                "kind": "Name",
                "value": "filter"
            },
            "value": {
                "kind": "ObjectValue",
                "fields": [
                    {
                        "kind": "ObjectField",
                        "name": {
                            "kind": "Name",
                            "value": "country"
                        },
                        "value": {
                            "kind": "StringValue",
                            "value": "US"
                        }
                    }
                ]
            }
        },
        {
            "kind": "Argument",
            "name": {
                "kind": "Name",
                "value": "options"
            },
            "value": {
                "kind": "ObjectValue",
                "fields": [
                    {
                        "kind": "ObjectField",
                        "name": {
                            "kind": "Name",
                            "value": "limit"
                        },
                        "value": {
                            "kind": "IntValue",
                            "value": "3"
                        }
                    }
                ]
            }
        }
    ],
    "selectionSet": {
        "kind": "SelectionSet",
        "selections": [
            {
                "kind": "Field",
                "name": {
                    "kind": "Name",
                    "value": "code"
                },
                "arguments": [],
                "directives": []
            },
            {
                "kind": "Field",
                "name": {
                    "kind": "Name",
                    "value": "city"
                },
                "arguments": [],
                "directives": []
            }
        ]
    }
}
```

Then using `print` on line 58 the input to `resolveGraphDBQuery` becomes:
```
{
  getNodeAirports(filter: {country: "US"}, options: {limit: 3}) {
    code
    city
  }
}
```

Then further along in `resolveGraphDBQuery` the query is transformed again using `gql` on line 1061 and we end up with this value for `obj`:
```
{
    "kind": "Document",
    "definitions": [
        {
            "kind": "OperationDefinition",
            "operation": "query",
            "variableDefinitions": [],
            "directives": [],
            "selectionSet": {
                "kind": "SelectionSet",
                "selections": [
                    {
                        "kind": "Field",
                        "name": {
                            "kind": "Name",
                            "value": "getNodeAirports"
                        },
                        "arguments": [
                            {
                                "kind": "Argument",
                                "name": {
                                    "kind": "Name",
                                    "value": "filter"
                                },
                                "value": {
                                    "kind": "ObjectValue",
                                    "fields": [
                                        {
                                            "kind": "ObjectField",
                                            "name": {
                                                "kind": "Name",
                                                "value": "country"
                                            },
                                            "value": {
                                                "kind": "StringValue",
                                                "value": "US",
                                                "block": false
                                            }
                                        }
                                    ]
                                }
                            },
                            {
                                "kind": "Argument",
                                "name": {
                                    "kind": "Name",
                                    "value": "options"
                                },
                                "value": {
                                    "kind": "ObjectValue",
                                    "fields": [
                                        {
                                            "kind": "ObjectField",
                                            "name": {
                                                "kind": "Name",
                                                "value": "limit"
                                            },
                                            "value": {
                                                "kind": "IntValue",
                                                "value": "3"
                                            }
                                        }
                                    ]
                                }
                            }
                        ],
                        "directives": [],
                        "selectionSet": {
                            "kind": "SelectionSet",
                            "selections": [
                                {
                                    "kind": "Field",
                                    "name": {
                                        "kind": "Name",
                                        "value": "code"
                                    },
                                    "arguments": [],
                                    "directives": []
                                },
                                {
                                    "kind": "Field",
                                    "name": {
                                        "kind": "Name",
                                        "value": "city"
                                    },
                                    "arguments": [],
                                    "directives": []
                                }
                            ]
                        }
                    }
                ]
            }
        }
    ],
    "loc": {
        "start": 0,
        "end": 105
    }
}
```

So my question would be whether there is a way to avoid having to transform the input this many times? I wonder if the open cypher query could even be built without having to transform the app sync event into a `gql` object (this might be a bigger change if it's possible).

kmcginnes: After digging in to the code a bit I think @andreachild is correct that we can make this a tad bit simpler in the short term.

The part that I think we can remove is the `print()` call and the `gql()` call. We are essentially doing this:

```
const obj = gql(print(selectionSet));
```

But the selectionSet has all the information we need in the format we need already. The only part that is missing is the wrapper around the selection set:

```json
{
    "kind": "Document",
    "definitions": [
        {
            "kind": "OperationDefinition",
            "operation": "query",
            "variableDefinitions": [],
            "directives": [],
            "selectionSet": /* <selectionSet goes here> */
        }
    ]
}
```

So we could probably just build up the object using JavaScript instead of all the gql parsing tricks.

```js
const obj = {
  kind: "Document",
  definitions: [
    {
      kind: "OperationDefinition",
      operation: "query",
      variableDefinitions: [],
      directives: [],
      selectionSet: opSelectionSet,
    },
  ],
};

const graphQuery = resolveGraphDBQuery(obj);
```

This will break the test code since they are passing a graphQL query directly. But that should be an easy transition:

```js
// before
const result = resolverModule.resolveGraphDBQuery(query.graphql);

// after
const result = resolverModule.resolveGraphDBQuery(gql(query.graphql));
```

@danielfinke What do you think?

danielfinke: @kmcginnes @andreachild yep, good suggestion. I considered doing this in the first place but I didn't want to change the public API for the module. I must have been tired that day - it was trivial to make it backwards-compatible. Now `resolveGraphDBQuery` can accept the AST directly, or still parse a query string passed in by callers that haven't been updated yet. Let me know what you both think

kmcginnes: @danielfinke Great catch on the public API part and making it backward compatible. I missed that. 

I'll let @andreachild give the final word on this PR, but I think I'm happy.

test/TestCases/Case01/Case01.04.test.js
```diff
@@ -4,7 +4,7 @@ import { readJSONFile, testResolverQueriesResults } from '../../testLib';
 
 const casetest = readJSONFile('./test/TestCases/Case01/case.json');
 
-await testResolverQueriesResults(   './TestCases/Case01/output/output.resolver.graphql.js',
```
andreachild, line 7: Thank you for fixing this! I know some of the queries in this test case fail but that's outside the scope of this PR.

test/TestCases/Case01/Case01.05.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+    test('should resolve queries with a filter', () => {
+        const result = resolve({
+            field: 'getNodeAirport',
+            arguments: { filter: { code: 'SEA' } },
+            selectionSetGraphQL: '{ city }'
+        });
+
+        expect(result).toEqual({
+            query: 'MATCH (getNodeAirport_Airport:`airport`{code: $getNodeAirport_Airport_code})\n' +
```
andreachild, line 20: This is great to be able to execute this logic check without having to actually execute the query. I will make use of this for my PR https://github.com/aws/amazon-neptune-for-graphql/pull/63 as well.

templates/JSResolverOCHTTPSTemplate.js
andreachild: Usually backwards compatibility for the module public API is not a concern since the utility deploys the resolver code alongside the `index.mjs` that references it in a zip for the lambda. But this backwards compatibility actually helps me with the work I'm doing to allow this utility to output Apollo server artifacts 👍

andreachild:
LGTM thanks for your patience and taking feedback into consideration

templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -65,6 +80,35 @@ const returnString = [];    // openCypher return statements
...
+}
+
+
+function getFieldDef(fieldName) {
+    const rootTypeDefs = getRootTypeDefs();
+
+    for (const rootDef of rootTypeDefs) {
+        const fieldDef = rootDef.fields.find(def => def.name.value === fieldName);
+
+        if (fieldDef) {
```
andreachild, line 105: Nit: although I could not trigger this scenario from app sync UI (there seems to be some validation before the lambda is triggered) it would be nice to have an error indication if the specified field was not found in the schema.

Cole-Greer (approved):
Thanks @danielfinke, LGTM

templates/JSResolverOCHTTPSTemplate.js
danielfinke: @andreachild as you observed, AppSync validates the schema ahead of time, but for good measure I have added default values according to AST spec and assertions for situations that should not occur: #69

PR #62, opened 2025-01-27
Don't cast integers to floats in Neptune schema

danielfinke:
*Issue #, if available:* #55

*Description of changes:*

Don't cast integers to floats in Neptune schema

Test case for cast in `Case04.02`

- update docs/samples/test cases

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #63, opened 2025-01-29
Retain numeric parameter value type when creating open cypher query 

andreachild:
*Issue #:* https://github.com/aws/amazon-neptune-for-graphql/issues/64
*Description of changes:*

Numeric graphQL query parameter values were inadvertently being set as strings upon conversion to the equivalent open cypher query. Fixed `transformFunctionInputParameters` which converts the graphQL request fields into a parameters object by converting the field value to a `Number` if the type is `IntValue` or `FloatValue`.

### DIFF
--- a/templates/JSResolverOCHTTPSTemplate.js
+++ b/templates/JSResolverOCHTTPSTemplate.js
@@ -24,7 +24,7 @@ const schemaDataModel = JSON.parse(schemaDataModelJSON);
 const schema = buildASTSchema(schemaDataModel, { assumeValidSDL: true });
 
 
-function resolveGraphDBQueryFromAppSyncEvent(event) {        
+function resolveGraphDBQueryFromAppSyncEvent(event) {
     const fieldDef = getFieldDef(event.field);
 
     const args = [];
@@ -67,8 +67,8 @@ function resolveGraphDBQueryFromAppSyncEvent(event) {
     const graphQuery = resolveGraphDBQuery(obj);
     return graphQuery;
 }
-  
-  
+
+
 function resolveGraphDBQueryFromApolloQueryEvent(event) {
   // TODO
 }
@@ -749,6 +749,9 @@ function transformFunctionInputParameters(fields, schemaInfo) {
         fields.forEach(field => {
             if (field.name.value === arg.name) {
                 let value = field.value.value;
+                if (field.value.kind === 'IntValue' || field.value.kind === 'FloatValue') {
+                    value = Number(value);
+                }
                 if (arg.name === schemaInfo.graphDBIdArgName) {
                     r.graphIdValue = value
                 } else if (arg.alias != null) {
--- a/test/TestCases/Case01/queries/Query0020.json
+++ b/test/TestCases/Case01/queries/Query0020.json
@@ -1,0 +1,18 @@
+{
+  "name": "Filter by parameter with numeric value and return mix of numeric value types",
+  "description": "",
+  "graphql": "query MyQuery {\n getNodeAirports(filter: { city: \"Seattle\", runways: 3 }) {\n code\n lat\n lon\n elev\n}\n }",
+  "parameters": {
+    "getNodeAirports_Airport_city": "Seattle",
+    "getNodeAirports_Airport_runways": 3
+  },
+  "resolved": "MATCH (getNodeAirports_Airport:`airport`{city: $getNodeAirports_Airport_city, runways: $getNodeAirports_Airport_runways})\nRETURN collect({code: getNodeAirports_Airport.`code`, lat: getNodeAirports_Airport.`lat`, lon: getNodeAirports_Airport.`lon`, elev: getNodeAirports_Airport.`elev`})",
+  "result": [
+    {
+      "code": "SEA",
+      "elev": 432,
+      "lon": -122.30899810791,
+      "lat": 47.4490013122559
+    }
+  ]
+}
--- a/test/TestCases/Case01/Case01.05.test.js
+++ b/test/TestCases/Case01/Case01.05.test.js
@@ -57,6 +57,22 @@ describe('AppSync resolver', () => {
         });
     });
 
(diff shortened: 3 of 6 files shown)


src/NeptuneSchema.js
```diff
@@ -202,8 +202,9 @@ function CastGraphQLType(value) {
     if (typeof value === 'number') {
         if (Number.isInteger(value)) {
             propertyType = 'Int';
+        } else {
+            propertyType = 'Float';
         }
-        propertyType = 'Float';
```
kmcginnes, line 206: ### NIT

This is totally subjective and outside the scope of your change. But I find that functions like this read better if they use the "return early" pattern, rather than the assign the result and return at the end.

Returning early allows your brain to scan the path of interest and stop. Assigning the result and return at the end forces you to read the entire function logic to make sure the value is not assigned elsewhere before being returned.

Here is what it would look like if it used return early:

```js
function CastGraphQLType(value) {
    if (typeof value === 'number') {
        return Number.isInteger(value)
            ? 'Int'
            : 'Float';
    }
    
    if (typeof value === 'boolean') {
        return 'Boolean';
    }

    if (value instanceof Date) {
        return 'Date';
    }
    
    return 'String';
}
```

templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -21,39 +21,45 @@ const schemaDataModelJSON = `INSERT SCHEMA DATA MODEL HERE`;
...
-    const graphQuery = resolveGraphDBQuery(query);
-    return graphQuery;
+
+    let argsStr = '';
+    if (queryArgs.length > 0) {
+        argsStr = `(${queryArgs.join(' ')})`;
+    }
+
+    return `{\n${event.field}${argsStr}${event.selectionSetGraphQL}\n}`;
+}
```
kmcginnes, line 53: ### NIT

Another set of small subjective quibbles. You can use `Object.entries()` to get both the key and value at the same time. You can use `map()` instead of `forEach()` to assign directly to `queryArgs` allowing you to make it `const`. And you can use ternary for `argsStr` to make that a `const`.

Can you tell that I'm a huge fan of `const`? 🤓

src/NeptuneSchema.js
andreachild: This change will go away since I'm going to rebase my branch once https://github.com/aws/amazon-neptune-for-graphql/pull/62/files is merged. The PR author has incorporated your suggestion I believe.

templates/JSResolverOCHTTPSTemplate.js
andreachild: Changed to use Object.entries and const

src/NeptuneSchema.js
andreachild: This change is now gone after rebase from main

templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -21,39 +21,64 @@ const schemaDataModelJSON = `INSERT SCHEMA DATA MODEL HERE`;
...
+ *
+ * @param event
+ * @returns {string}
+ */
+function appSyncEventToGraphQLQuery(event) {
+    const queryArgs = Object.entries(event.arguments).map(([key, value]) => {
+        if (typeof value === 'object') {
+            let nestedArgs = Object.entries(value)
+                .map(([nestedKey, nestedValue]) => `${nestedKey}: ${addQuotesIfString(nestedValue)}`)
+                .join(', ');
```
kmcginnes, line 58: Is one layer of nesting all we need to support?

Could an argument be something like:

```
filter: { country: 'US', runways: 7, location: { city: 'Vancouver' } }
```

test/TestCases/Case01/queries/Query0020.json
kmcginnes, line 1: Does this need to be referenced anywhere, or is it automatically found and tested against?

kmcginnes (commented):
Looks good. Just a couple of questions

test/TestCases/Case01/queries/Query0020.json
andreachild: It's automatically detected by the code in `Case01.03.test.js` which references the queries directory:

```
await testResolverQueries('./TestCases/Case01/output/output.resolver.graphql.cjs', './test/TestCases/Case01/queries');
```

templates/JSResolverOCHTTPSTemplate.js
danielfinke: @kmcginnes might be worth replacing this with my implementation in #61. It will handle nested objects and really anything that can be specified by a GraphQL input type

andreachild: Technically it is possible with graphql queries in general. However for this utility the query input arguments are built from the discovered node properties of the neptune graph and neptune [does not support complex properties](https://docs.aws.amazon.com/neptune/latest/userguide/access-graph-opencypher-data-model.html) (yet).

> Below is the list of data types supported in this release:
> 
> Bool
> Byte
> Short
> Int
> Long
> Float (Includes plus and minus Infinity and NaN, but not INF)
> Double (Includes plus and minus Infinity and NaN, but not INF)
> DateTime
> String

@danielfinke I agree your PR could help solve the numeric type value issue I just had a [concern](https://github.com/aws/amazon-neptune-for-graphql/pull/61#discussion_r1936314330) about the transformations. If it could be addressed then it's possible I can use your solution with the additional change I have in line 733 to fix the numeric type issue.

kmcginnes (approved):
Approving this smaller fix while we further investigate @danielfinke's more encompassing changes.

templates/JSResolverOCHTTPSTemplate.js
andreachild: I have rebased my branch off of main after https://github.com/aws/amazon-neptune-for-graphql/pull/61 was merged.

Cole-Greer (approved):
Thanks @andreachild, LGTM.

Issue #65, opened 2025-01-30
Backticks in input GraphQL schema file break resolver output


danielfinke:
If you use backticks in an input GraphQL schema file, such as in the following field and type descriptions,

```graphql
# Comment test with "double quotes" and 'single quotes'

"Members of `Group`s"
type User {
  firstName: String
  lastName: String
  "Computed based on `firstName` and `lastName`"
  fullName: String
}

"Description test with 'single quotes'"
type Group {
  name: String
}
```

the output of `schemaDataModelJSON` in the resolver output will be badly formed and the module will break.

Backticks in comments are fine because they are already stripped from the schema. The example above includes other JS string delimiters to account for in the fix.

Command: `node ./neptune-for-graphql.js --input-schema-file test.graphql --output-no-lambda-zip`

PR #66, opened 2025-01-30
Improve string delimiter escaping in `schemaDataModelJSON`

danielfinke:
*Issue #, if available:* #65

*Description of changes:*

Wrap the JSON in single quotes instead of backticks to reduce the escape sequence surface area. Escape any single quotes in the stringified JSON. This handles single and double quotes and backticks in the input schema's scalar, field/type/etc. descriptions.

- add test `Case03.02` to check output for schema with string delimiters
- update other `outputReference` resolvers
- add templates to lint check
  - fix resulting ESLint errors in templates
- exclude output from ESLint rules (hide errors for those files when using editor integrations)

copied changes from aws/amazon-neptune-for-graphql#61:

- fix query result tests that were sensitive to object key order (use `toEqual` instead of strict JSON string equality)
- also check output file size of `.cjs` resolver in test `Case01.02`
- fix running test `Case01.04` (use `.cjs` extension)


andreachild:
Hi @danielfinke hoping to review this sometime this week

src/resolverJS.js
```diff
@@ -15,12 +15,13 @@ import { loggerError } from "./logger.js";
 
 function resolverJS (schemaModel, queryLanguage, queryClient, __dirname) {
     let code = '';
-    const queryDataModelJSON = JSON.stringify(schemaModel, null, 2);
```
Cole-Greer, line 18: Hi @danielfinke, I wanted to ask about the reasoning for dropping the space argument from the call to `JSON.stringify()`. I think there is some value lost in the legibility of the formatted output.

andreachild:
@danielfinke I am curious about the driving factor of this PR. Is it to resolve ESLint issues or did you find a scenario which produced a graphQL schema that wasn't working correctly with the current escaping logic?

src/resolverJS.js
danielfinke: @Cole-Greer this is a caveat of using `'` as the delimiter: multi-line strings are not available

danielfinke:
@andreachild It was for a schema that would break the module - the JS syntax would become invalid. I have an example schema in #65.

@andreachild @Cole-Greer I think a better optional overall, if it acceptable, would be writing the schema data model JSON to a bundled `.json` file that is read by the resolver using `fs.readFile`. I haven't thought about any issues with that yet. Maybe a follow-up issue after this one is merged?

andreachild:
> @andreachild @Cole-Greer I think a better optional overall, if it acceptable, would be writing the schema data model JSON to a bundled `.json` file that is read by the resolver using `fs.readFile`. I haven't thought about any issues with that yet. Maybe a follow-up issue after this one is merged?

I very much like your idea of moving the schema to its own file and would prefer to skip changing the escaping logic as an intermediate step.

PR #69, opened 2025-02-10
Add assertions and type safety after #61

danielfinke:
*Issue #, if available:* relates to #61 (see https://github.com/aws/amazon-neptune-for-graphql/pull/61#discussion_r1937773379)

*Description of changes:*

Also fix mistyped return annotation on `resolveGraphDBQuery`, my bad

### DIFF
--- a/templates/JSResolverOCHTTPSTemplate.js
+++ b/templates/JSResolverOCHTTPSTemplate.js
@@ -10,7 +10,8 @@ express or implied. See the License for the specific language governing
 permissions and limitations under the License.
 */
 
-import { astFromValue, buildASTSchema, typeFromAST } from 'graphql';
+import assert from 'node:assert';
+import { astFromValue, buildASTSchema, isInputType, typeFromAST } from 'graphql';
 import { gql } from 'graphql-tag'; // GraphQL library to parse the GraphQL query
 
 const useCallSubquery = false;
@@ -27,12 +28,17 @@ const schema = buildASTSchema(schemaDataModel, { assumeValidSDL: true });
 export function resolveGraphDBQueryFromAppSyncEvent(event) {
     const fieldDef = getFieldDef(event.field);
 
+    assert(fieldDef);
+
     const args = [];
-    for (const inputDef of fieldDef.arguments) {
+    for (const inputDef of fieldDef.arguments ?? []) {
         const value = event.arguments[inputDef.name.value];
 
         if (value) {
             const inputType = typeFromAST(schema, inputDef.type);
+
+            assert(isInputType(inputType));
+
             args.push({
                 kind: 'Argument',
                 name: { kind: 'Name', value: inputDef.name.value },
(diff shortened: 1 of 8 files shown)


test/TestCases/Case05/case01.json
```diff
@@ -9,7 +9,5 @@
             "--create-update-aws-pipeline-neptune-endpoint", "<AIR_ROUTES_DB_HOST>:<AIR_ROUTES_DB_PORT>",
             "--output-resolver-query-https"],
     "host": "<AIR_ROUTES_DB_HOST>",
-    "port": "<AIR_ROUTES_DB_PORT>",
-    "testOutputFilesSize": ["output.resolver.graphql.js", "output.schema.graphql", "output.source.schema.graphql"],
```
andreachild, line 13: Since you're cleaning up test files with unused fields can you also do the same for `Case03/case.json`?

templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -27,12 +28,17 @@ const schema = buildASTSchema(schemaDataModel, { assumeValidSDL: true });
 export function resolveGraphDBQueryFromAppSyncEvent(event) {
     const fieldDef = getFieldDef(event.field);
 
+    assert(fieldDef);
```
andreachild, line 31: For troubleshooting would be nice to add a value for the optional `message` parameter to indicate what expected state was not satisfied ie. `Field definition for ${event.field} not found`.

templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -27,12 +28,17 @@ const schema = buildASTSchema(schemaDataModel, { assumeValidSDL: true });
...
+
     const args = [];
-    for (const inputDef of fieldDef.arguments) {
+    for (const inputDef of fieldDef.arguments ?? []) {
         const value = event.arguments[inputDef.name.value];
 
         if (value) {
             const inputType = typeFromAST(schema, inputDef.type);
+
+            assert(isInputType(inputType));
```
andreachild, line 40: For troubleshooting would be nice to add a value for the optional `message` parameter to indicate what expected state was not satisfied ie. `Type ${inputType} is not a valid input type`.

templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -99,8 +105,8 @@ function getTypeDefs(typeNameOrNames) {
 function getFieldDef(fieldName) {
     const rootTypeDefs = getRootTypeDefs();
 
-    for (const rootDef of rootTypeDefs) {
-        const fieldDef = rootDef.fields.find(def => def.name.value === fieldName);
+    for (const rootTypeDef of rootTypeDefs) {
+        const fieldDef = rootTypeDef.fields?.find(def => def.name.value === fieldName);
```
andreachild, line 109: Perhaps the `Case03` which is not currently asserting any validation after the utility is executed can be repurposed to verify error handling for scenarios addressed in this PR (app sync event fields do not match up with the graphQL schema). This could be done by using similar approach to your `Case01.05.test.js` tests except modifying the app sync event to reference unknown types.

Issue #67, opened 2025-02-04
Support for output of Apollo Server artifacts


andreachild:
For users that are using Apollo Server instead of AWS App Sync and Lambda it would be nice to be able to specify an option to indicate that the utility should output a ZIP of artifacts that can be deployed with Apollo Server.

Example of usage could be:

```
--input-graphdb-schema-neptune-endpoint
abc.us-west-2.neptune.amazonaws.com:8182
--create-update-apollo-server
--output-resolver-query-https
```

If the user wants to use Apollo Server subgraph they could instead use:

```
--create-update-apollo-server-subgraph
```

Which would then create a ZIP file in the output folder which contains artifacts that can be deployed via Apollo Server:

```
package.json
package-lock.json
.env
output.resolver.graphql.js
output.schema.graphql
main.js
<<other supporting js files>>
```

To deploy the artifacts locally the workflow would be:
1. unzip the output file
2. execute `npm install` in the unzipped folder
3. execute `node main.js`

This would then start the Apollo Server on a local environment:

![Image](https://github.com/user-attachments/assets/4ef35316-68c3-43ee-b235-e368dd82fa84)

andreachild:
Done by https://github.com/aws/amazon-neptune-for-graphql/pull/70

PR #70, opened 2025-02-12
Support output of zip package of Apollo Server artifacts

andreachild:
*Issue #, if available:* https://github.com/aws/amazon-neptune-for-graphql/issues/67

*Description of changes:*

Added options `--create-update-apollo-server` and `--create-update-apollo-server-subgraph` which when specified will cause the utility to create a zip file of Apollo Server artifacts instead of AWS App Sync artifacts. After the zip is unpackaged the contents can then be used to deploy a GraphQL API to interface with neptune db or neptune analytics via Apollo Server.

Limitations: mutations not yet supported, only http queries to neptune are supported (not sdk)

To support this feature the following additional modifications were made:
* extracted logic which sends an http request to neptune into a new template file `queryHttpNeptune.mjs` so that it can be reused for lambda and Apollo
* renamed `lambdaZip.js` to `zipPackage.js` and added function for creating an Apollo package
* added new template folder for Apollo related files
* fixed bug with resolver template that caused it to fail if there are no query arguments specified
* added new environment-agnostic resolver function `resolveGraphDBQueryFromEvent` which can take a selection set that is already an AST object as opposed to a string
* fixed bug with lambda zip packaging which didn't account for customized resolver file paths
* upgraded axios version as previous version had a vulnerability


kmcginnes:
Since this tool does a lot of file system related things, I wanted to recommend [memfs](https://github.com/streamich/memfs) for testing file system logic. It allows your file system tests to be way less fragile at the small cost of a bit of pain setting it up initially. 

I use it in Graph Explorer to test the server code, which deals with paths and checking for file existence.

src/zipPackage.js
```diff
@@ -0,0 +1,74 @@
...
+        archive.file(resolverFilePath, {name: 'output.resolver.graphql.js'})
+        if (options?.http) {
+            const modulePath = getModulePath();
+            archive.file(modulePath + '/../templates/queryHttpNeptune.mjs', {name: 'queryHttpNeptune.mjs'})
+        }
+        await archive.finalize();
+    } catch (error) {
+        loggerError('Lambda deployment package creation failed', error);
+    }
+}
```
kmcginnes, line 36: Something about this abstraction is not sitting right with me. I can see the patterns of reuse that I imagine you also see. A majority of this code is the same in every use case. But there is still a fraction of it that requires flexibility, hence your new parameters.

I wonder if we can approach this differently. If I look at this from the perspective of the way it is used now, there are two points of customization:

1. `resolverFilePath`
2. Additional file needed for http

Both of these are related to adding files to the archive. So, what if the call site for `createLambdaDeploymentPackage` was in control of the files added, but not in control of the archive creation. I think this would give you the reduction in code duplication you desire and allow the flexibility needed for all situations. It'll also make it clear what those customization points are doing at the call site, rather than as conditional logic within the abstraction.

So it might look something like this:

```js
await createLambdaDeploymentPackage(
  templateFolderPath, 
  zipFilePath,
  [
    { source: resolverFilePath, target: 'output.resolver.graphql.js' },
    { 
      source: Path.join(getModulePath(), '..', 'templates', queryHttpNeptune.mjs),
      target: 'queryHttpNeptune.mjs'
    }
  ]
);
```

I'd even go further and add helper functions to build those individual file types. So if we had the helpers it would reduce down to:

```js
await createLambdaDeploymentPackage(
  templateFolderPath, 
  zipFilePath,
  [
    createResolverArchiveFile(resolverFilePath),
    createHttpArchiveFile()
  ]
);
```

And if you wanted to go all the way by naming the parameters so they are easier to understand:

```js
await createLambdaDeploymentPackage({
  templateFolder: templateFolderPath, 
  outputFile: zipFilePath,
  archiveFiles: [
    createResolverArchiveFile(resolverFilePath),
    createHttpArchiveFile()
  ]
});
```

templates/ApolloServer/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "name": "neptune-apollo-server",
+  "version": "1.0.0",
+  "description": "",
+  "license": "Apache-2.0",
+  "author": "",
+  "type": "module",
+  "main": "index.mjs",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
```
kmcginnes, line 11: NIT: you can just delete the `scripts` section if it is not needed

templates/queryHttpNeptune.mjs
```diff
@@ -0,0 +1,77 @@
...
+                gremlin: resolvedQuery.query
+            }, requestConfig);
+        }
+        if (loggingEnabled) {
+            console.log("Query result: ", JSON.stringify(response.data, null, 2));
+        }
+
+        if (resolvedQuery.language === 'gremlin') {
+            const gremlinData = response.data["result"]["data"];
+            const jsonData = refactorGremlinqueryOutput(gremlinData, resolvedQuery.fieldsAlias);
```
kmcginnes, line 51: This is not a new problem to this PR, but the response structure from Neptune will depend on the version of Neptune, or more specifically the version of Gremlin Server.

Neptune 1.3.2.0 introduced Gremlin 3.7, which changes the default JSON serialized structure. So before 1.3.2.0 you would map the JSON data one way, after you would need to map a different way.

I worked around this by specifying the serialization method in a header, but this is probably not the best approach since Gremlin 4 will remove support for the serialization method I'm using.

templates/queryHttpNeptune.mjs
```diff
@@ -0,0 +1,77 @@
...
+            if (loggingEnabled) {
+                console.log("Gremlin query json data: ", jsonData);
+            }
+            return JSON.parse(jsonData);
+        }
+
+        if (resolvedQuery.language === 'opencypher') {
+            let openCypherData = response.data;
+            if (openCypherData.results.length === 0) {
+                return null;
```
kmcginnes, line 61: Should this be an error, or at the very least log out the response data you did get?

src/main.js
```diff
@@ -578,7 +623,7 @@ async function main() {
 
             try {
                 if (!quiet) spinner = ora('Creating Lambda ZIP ...').start();
-                await createLambdaDeploymentPackage(__dirname + outputLambdaPackagePath, outputLambdaResolverZipFile);                                
+                await createLambdaDeploymentPackage(__dirname + outputLambdaPackagePath, outputLambdaResolverZipFile, outputLambdaResolverFile, {http: queryClient === 'http'});
```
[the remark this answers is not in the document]
andreachild: `outputLambdaResolverFile` is actually a path to a js file not a folder and if the value is empty then it's given a default value earlier in the logic:

```
// Output Lambda resolver
if (outputLambdaResolverFile == '') { 
    outputLambdaResolverFile = outputFolderPath + '/output.resolver.graphql.js'; 
}
```

README.md
```diff
@@ -271,15 +272,33 @@ See CDK end to end example [here](https://github.com/aws/amazon-neptune-for-grap
 [Here](https://github.com/aws/amazon-neptune-for-graphql/blob/main/doc/resources.md) the detailed list of resources needed to configure the GraphQL API pipeline.
 <br>
 
+# Generate Apollo Server Artifacts
+If you prefer to use Apollo Server instead of AWS App Sync, the utility can generate a ZIP file of Apollo Server artifacts with the CLI option `--create-update-apollo-server` for a standalone server or `--create-update-apollo-server-subgraph` for federated systems.
+
+For example:
+`neptune-for-graphql --input-graphdb-schema-neptune-endpoint
+abc.us-west-2.neptune.amazonaws.com:8182 --create-update-apollo-server --output-resolver-query-https`
```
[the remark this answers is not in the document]
andreachild: Changed as suggested

src/main.js
```diff
@@ -568,6 +598,21 @@ async function main() {
             break;
         }
 
+        // output Apollo zip
+        if (createUpdateApolloServer || createUpdateApolloServerSubgraph) {
+            const apolloZipPath = path.join(outputFolderPath, 'apollo.server.zip');
+            try {
+                if (!quiet) spinner = ora('Creating Apollo server ZIP file ...').start();
```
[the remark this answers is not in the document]
andreachild: Changed to fail the spinner on exception

README.md
```diff
@@ -271,15 +272,33 @@ See CDK end to end example [here](https://github.com/aws/amazon-neptune-for-grap
...
+
+The command above will generate an `apollo-server.zip` file which can then be deployed locally by following these steps:
+1. unzip `apollo-server.zip`
+2. change directory into the unzipped folder
+3. execute `npm install` to install required dependencies
+4. execute `node index.mjs` to start the Apollo Server
+5. access the graphQL application in a browser by visiting `http://localhost:4000/`
+
+Note that Node's default AWS credentials provider is used for authentication with Neptune. See [AWS SDK credential providers](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromnodeproviderchain) for more information.
```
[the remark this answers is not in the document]
andreachild: Changed to use `[!NOTE]`

src/zipPackage.js
andreachild: I have refactored the zip related functions to separate the zip file creation from logic that determines what goes inside the zip.

templates/ApolloServer/package.json
andreachild: I think I will keep it since It seems to be a commonly used mechanism in this project and some javascript projects to ensure that whoever is running `npm test` is made aware that there are actually no written tests.

templates/queryHttpNeptune.mjs
andreachild: After debugging I found this is not an error case but occurs when there was a query with a match clause and limit 1 (meaning the utility is attempting to fetch a single entity) but no result found. In this case the result should be null. I added a comment to clarify it is not an error.

andreachild:
> Since this tool does a lot of file system related things, I wanted to recommend [memfs](https://github.com/streamich/memfs) for testing file system logic. It allows your file system tests to be way less fragile at the small cost of a bit of pain setting it up initially.
> 
> I use it in Graph Explorer to test the server code, which deals with paths and checking for file existence.

I think this is a good idea but would like to explore this in a separate PR.

templates/queryHttpNeptune.mjs
andreachild: I'll have to see if I can find out what versions of neptune this utility is meant to support but for now will keep the current logic.

src/zipPackage.js
kmcginnes: Looks good

kmcginnes (approved):
LGTM

PR #78, opened 2025-02-28
Configure GitHub to run unit tests for PRs and commits to main

flora-jin:
Created a github action that is triggered by PRs and commits to main that executes unit tests. Action fails if unit tests fail and passes if unit tests pass.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -1,0 +1,32 @@
+name: Unit Tests
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  install-and-test:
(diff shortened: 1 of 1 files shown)


.github/workflows/unit.yml
```diff
@@ -0,0 +1,32 @@
...
+    branches:
+      - main
+
+jobs:
+  install-and-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version:
+          - 23.9.0
```
andreachild, line 18: For consistency with the [npm publish workflow](https://github.com/aws/amazon-neptune-for-graphql/blob/main/.github/workflows/npm-publish.yml) should change to version 20

kmcginnes (approved):

PR #79, opened 2025-03-12
Separate graphQL schema from resolver

flora-jin:
The current graphQL utility has a resolver template file with a placeholder for the graphQL schema. During runtime, the utility replaces the placeholder with the schema and then includes the resulting resolver schema in the zip contents that are deployed to AWS lambda. 

However, this logic makes it difficult to unit test so the pull request includes code changes where the schema is separated from the resolver. Instead of the resolver template having a placeholder for the schema, it instead takes in the schema as a separate file that is deployed with the other files in the zip file.


templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -12,13 +12,14 @@ permissions and limitations under the License.
...
 import { gql } from 'graphql-tag'; // GraphQL library to parse the GraphQL query
+import {readFileSync} from "fs";
 
 const useCallSubquery = false;
 
 // TIMESTAMP HERE
 
-const schemaDataModelJSON = `INSERT SCHEMA DATA MODEL HERE`;
-    
+let schemaDataModelJSON = readFileSync('resolver.schema.graphql', 'utf-8');
```
acarbonetto, line 21:
```suggestion
const schemaDataModelJSON = readFileSync('resolver.schema.graphql', 'utf-8');
```

doc/todoExample.md
```diff
@@ -21,7 +21,7 @@ Let's now run this schema through the utility and create the GraphQL API in AWS
 
 `neptune-for-graphql --input-schema-file ./samples/todo.schema.graphql --create-update-aws-pipeline --create-update-aws-pipeline-name TodoExample --create-update-aws-pipeline-neptune-endpoint` <*your-neptune-database-endpoint:port*> ` --output-resolver-query-https`
 
-Te utility created a new file in the *output* folder called *TodoExample.source.graphql*, and the GraphQL API in AppSync. As you can see below, the utility inferenced:
+The utility created a new file in the *output* folder called *TodoExample.source.graphql*, and the GraphQL API in AppSync. As you can see below, the utility inferenced:
```
acarbonetto, line 24: 🫶

src/resolverJS.js
```diff
@@ -13,15 +13,13 @@ permissions and limitations under the License.
...
 
-function resolverJS (schemaModel, queryLanguage, queryClient, __dirname) {
+function resolverJS (queryLanguage, queryClient, __dirname) {
     let code = '';
-    const queryDataModelJSON = JSON.stringify(schemaModel, null, 2);
     
     if (queryLanguage == 'opencypher') {
         try {
             code = readFileSync(__dirname + '/../templates/JSResolverOCHTTPSTemplate.js');
             code = code.toString().replace('TIMESTAMP HERE', (new Date()).toISOString());
```
andreachild, line 22: Since we are changing the resolver to no longer be a 'template' I don't think we need the timestamp anymore and it can be removed entirely from the resolver template.

src/main.js
```diff
@@ -519,11 +519,21 @@ async function main() {
         // Validate schema
         schemaModel = validatedSchemaModel(schemaModel, quiet);
 
+        // Generate schema for resolver
+        const queryDataModelJSON = JSON.stringify(schemaModel, null, 2);
+        let resolverSchemaFile = `${outputFolderPath}/resolver.schema.graphql`
```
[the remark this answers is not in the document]
flora-jin: When I use the createUpdatePipelineName within the file format, how do I obtain the file name in JSResolverOCHTTPSTemplate.js in order to read the file ? I couldn't think of another way to obtain the file name without asking for it dynamically in the resolverJS.js file.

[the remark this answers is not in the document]
andreachild: The file that is created in the output folder can have a name that depends on `createUpdatePipelineName`. The equivalent file that is added to the ZIP that is used to deploy the lambda can be a static name that never changes. See `zipPackage.js` the function `createApolloDeploymentPackage` has an example of this.

src/zipPackage.js
```diff
@@ -50,6 +50,11 @@ export async function createLambdaDeploymentPackage({outputZipFilePath, template
             source: path.join(getModulePath(), '/../templates/queryHttpNeptune.mjs')
         })
     }
+
+    filePaths.push({
+        source: path.join(getModulePath(), '/../output/resolver.schema.graphql')
```
andreachild, line 55: This will not work if the user specifies an alternative output folder via the `--output-folder-path` argument because this code assumes that the output folder is always 'output' (the default value).

templates/JSResolverOCHTTPSTemplate.js
```diff
@@ -12,13 +12,14 @@ permissions and limitations under the License.
...
 import { gql } from 'graphql-tag'; // GraphQL library to parse the GraphQL query
+import {readFileSync} from "fs";
 
 const useCallSubquery = false;
 
 // TIMESTAMP HERE
 
-const schemaDataModelJSON = `INSERT SCHEMA DATA MODEL HERE`;
-    
+const schemaDataModelJSON = readFileSync('resolver.schema.graphql', 'utf-8');
```
andreachild, line 21: This change works when executing the lambda since the `resolver.schema.graphql` file is always packaged with the `index.js` file in the same directory, however since it assumes that the `resolver.schema.graphql` file is in the current working directory (the location that the node process is executing from) it causes the integration tests to fail as the file is not in the working directory when the tests are being executed.

My suggestion would be to change the resolver to require callers to 'initialize' the resolver with the location of the relevant schema file. This way the location of the schema file is not hardcoded and more flexible.

It could look something like this:

```
let schemaDataModel;
let schema;
                                   
export function initSchema(schemaFilePath) {
    const schemaDataModelJSON = readFileSync(schemaFilePath, 'utf-8');
    schemaDataModel = JSON.parse(schemaDataModelJSON);
    schema = buildASTSchema(schemaDataModel, { assumeValidSDL: true });
}
```

Then the callers need to do this:

```
initSchema(<<PATH_TO_RESOLVER_SCHEMA>>);
const resolver = resolveGraphDBQueryFromAppSyncEvent(event);
```

src/resolverJS.js
andreachild, line 1: I don't think this file is useful anymore since it's not replacing anything in the resolver. Its usage can be replaced by a call to `readFileSync` directly instead.

src/main.js
andreachild, line 1: Since the resolver is no longer a 'template' file as its content is no longer dynamic, we shouldn't need to save the resolver file as an output any longer. Therefore the following code should no longer be necessary:

```
writeFileSync(outputLambdaResolverFile, outputLambdaResolver);
...
writeFileSync(outputJSResolverFile, outputJSResolver);
```

Instead whatever is referencing these dynamic output resolver files should just reference the static resolver file instead.

src/zipPackage.js
```diff
@@ -82,6 +90,8 @@ export async function createApolloDeploymentPackage({zipFilePath, resolverFilePa
             {source: path.join(modulePath, '/../templates/ApolloServer')},
             {source: resolverFilePath, target: 'output.resolver.graphql.js'},
             {source: schemaFilePath, target: 'output.schema.graphql'},
+            {source: resolverSchemaFilePath, target: 'resolver.schema.json'},
```
andreachild, line 93: Please add the `output.` prefix for consistency with the other files.

templates/JSResolverOCHTTPSTemplate.js
andreachild, line 1: Since this is no longer a template file, it can be renamed to `JSResolverOCHTTPS.js`.

templates/Lambda4AppSyncGraphSDK/index.mjs
```diff
@@ -49,6 +49,7 @@ function log(message) {
  */
 function resolveGraphQuery(event) {
     try {
+        initSchema('resolver.schema.json');
```
andreachild, line 52: A similar change is needed in the `ApolloServer/neptune.mjs` file.

src/main.js
flora-jin: Too big of a refactoring for this ticket so will leave out for now.

src/main.js
```diff
@@ -519,11 +518,40 @@ async function main() {
         // Validate schema
         schemaModel = validatedSchemaModel(schemaModel, quiet);
 
-        // Generate AWS Lambda resolver for AppSync and Amazon Neptune        
-        outputLambdaResolver = resolverJS(schemaModel, queryLanguage, queryClient, __dirname);
+        // Generate schema for resolver
+        const queryDataModelJSON = JSON.stringify(schemaModel, null, 2);
+
+        let resolverSchemaFile;
+        if (createUpdatePipelineName == '') {
```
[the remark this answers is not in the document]
flora-jin: At the top of main.js it is initialized as createUpdatePipelineName = '' and stays that way throughout the code if the user does not input a name for the pipeline.

src/zipPackage.js
```diff
@@ -62,11 +69,12 @@ export async function createLambdaDeploymentPackage({outputZipFilePath, template
  * @param zipFilePath the file path where the zip should be created
  * @param resolverFilePath path to the resolver file to include in the zip
  * @param schemaFilePath path to the schema file to include in the zip
+ * @param resolverSchemaFilePath path to the resolver schema file to include in the zip
```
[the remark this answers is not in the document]
flora-jin: The schemaFilePath and resolverSchemaFilePath are two different file paths as the schemaFilePath leads to the output.schema.graphql file which is the actual schema file and resolverSchemaFilePath leads to the output.resolver.schema.json file which is the parsed schema file to be inputted into the resolver.

Cole-Greer (approved):

PR #83, opened 2025-04-03
Added resolver unit tests

flora-jin:
After separating the resolver from the schema, the resolver is able to now have its own unit tests aside from the integration tests.

- created a new unit test file, _resolver.test.js_  that holds tests that have graphQL query input and validate expected open cypher query output
- moved the resolver tests from Case01.05.test.js into the new unit test file
- added the resolver query tests from TestCases/Case01/queries to the new unit test file
- added additional unit tests for extra queries specifically:
      - query with no parameters
      - query with parameters that have constant values
* did not add separate queries for:
      - query with parameters that have variable values as it has the same event as query with parameters as constant values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,3 +39,4 @@ This release contains new support for Apollo Server integration.
 * Changed lambda template to use ECMAScripts modules (#68)
 * Add template file missing from packaging (#71)
 * Separated graphQL schema from resolver template (#79)
+* Added unit tests for resolver and moved resolver integration tests to be unit tests (#83)
--- a/src/test/airports.customized.graphql
+++ b/src/test/airports.customized.graphql
@@ -1,0 +1,132 @@
+type Continent @alias(property:"continent") {
+_id: ID! @id
+type: String
+code: String
+desc: String
+airportContainssOut(filter: AirportInput, options: Options): [Airport] @relationship(edgeType:"contains", direction:OUT)
+contains:Contains
+}
+
+input ContinentInput {
+_id: ID @id
+type: String
+code: String
+desc: String
+}
+
+type Country @alias(property:"country") {
+_id: ID! @id
+type: String
+code: String
+desc: String
+airportContainssOut(filter: AirportInput, options: Options): [Airport] @relationship(edgeType:"contains", direction:OUT)
+contains:Contains
+}
+
+input CountryInput {
+_id: ID @id
+type: String
+code: String
+desc: String
+}
+
+type Version @alias(property:"version") {
+_id: ID! @id
+date: String
+desc: String
+author: String
+type: String
+code: String
+}
+
+input VersionInput {
+_id: ID @id
+date: String
+desc: String
+author: String
+type: String
+code: String
+}
+
+type Airport @alias(property:"airport") {
+_id: ID! @id
+type: String
+city: String
+icao: String
+code: String
+country: String
+lat: Float
+longest: Int
+runways: Int
+desc: String
+lon: Float
+region: String
+elev: Int
+outboundRoutesCount: Int @graphQuery(statement: "MATCH (this)-[r:route]->(a) RETURN count(r)")
+continentContainsIn: Continent @relationship(edgeType:"contains", direction:IN)
+countryContainsIn: Country @relationship(edgeType:"contains", direction:IN)
+airportRoutesOut(filter: AirportInput, options: Options): [Airport] @relationship(edgeType:"route", direction:OUT)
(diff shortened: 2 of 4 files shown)


src/test/resolver.test.js
```diff
@@ -0,0 +1,497 @@
+import { initSchema, resolveGraphDBQueryFromAppSyncEvent, resolveGraphDBQuery } from './output.resolver.graphql.js'
```
andreachild, line 1: You should reference `JSResolverOCHTTPS.js` instead of adding test specific resolver file. Otherwise if someone updates the `JSResolverOCHTTPS.js` file they must also remember to update the `output.resolver.graphql.js` file in the test folder to check that the change didn't break anything.

src/test/queries/Query0001.json
```diff
@@ -0,0 +1,10 @@
+{  
+    "name": "getAirport",
+    "description": "Inference query from return type",
+    "graphql": "query MyQuery {\n getAirport(code: \"SEA\") {\n city \n }\n}",
```
andreachild, line 4: Instead of having one query per file these queries should be inlined in the `resolver.test.js` file as opposed to reading each query from a file.

src/test/resolver.test.js
```diff
@@ -0,0 +1,497 @@
...
+import fs, { readFileSync } from "fs";
+import gql from "graphql-tag";
+
+// Initialize resolver
+const schemaDataModelJSON = readFileSync('src/test/output.resolver.schema.json', 'utf-8');
```
andreachild, line 6: You should be able to use the existing schema file `airports.graphql` in the test directory so we don't add more copies of the same schema unnecessarily. You'll have to modify the test queries to match the schema or the schema to match the queries. Something like this:

```
// Initialize resolver
const airportSchema = readFileSync('src/test/airports.graphql', 'utf-8');
const airportAST = gql`
    ${airportSchema}
`;
let schemaModel = JSON.stringify(airportAST, null, 2);
schemaModel = JSON.parse(schemaModel); // don't know why this JSON handling is needed but won't work otherwise
initSchema(schemaModel);
```

flora-jin: As it is hard to fix the airports.graphql test in graphdb.test.js of comparing to the airports-neptune-schema.json, instead created new schema file for resolver.test.js called airports.customized.graphql that is a copy of airports.graphql with extra queries

src/test/resolver.test.js
andreachild, line 1: It is conventional to name the test file the same as the js file being tested. In this case it should be called `JSResolverOCHTTPS.test.js`

src/test/resolver.test.js
```diff
@@ -0,0 +1,490 @@
...
+        parameters: {},
+        language: 'gremlin',
+        refactorOutput: null
+    });
+});
+
+// Resolver Query Tests
+
+// Query0001
+test('should return correct query for Query0001: getAirport', () => {
```
andreachild, line 144: It would be more informative if the test titles referenced the wording from the original integration test description. For example in this case the Query0001 description is 'Inference query from return type' so the test title could be something like 'should inference query from return type'.

src/test/resolver.test.js
```diff
@@ -0,0 +1,490 @@
...
+            getNodeAirports_Airport_city: 'Seattle',
+            getNodeAirports_Airport_runways: 3
+        },
+        language: 'opencypher',
+        refactorOutput: null
+    });
+});
+
+// Query0021
+test('should return correct query for Query0021: Query with no parameters', () => {
```
andreachild, line 464: I believe query 21 and 22 are 'new' tests and not part of the integration test suite. In that case we don't need to carry on incrementing the query number as it doesn't reference an integration test. It should be sufficient for the test title to describe what the test is validating.

src/test/resolver.test.js
flora-jin: Just to clarify, are you suggesting that I move the JSResolverOCHTTPS.test.js file into a new sub directory test under the templates folder. Also, do I move the airports.customized.graphql to this new test folder as well.

andreachild: I was suggesting the existing test folder should have a new subfolder for templates and the JSResolverOCHTTPS.test.js file should be moved there. The airports.customized.graphql is OK where it is as it should stay with the other test graphql schema files.

```
test
--airports.customized.graphql
--templates
----JSResolverOCHTTPS.test.js
```

andreachild:
Please add a changelog entry now that we have a changelog 😄

andreachild:
LGTM thanks @flora-jin

Issue #86, opened 2025-05-02
Potential out of memory error retrieving edge directions for large graphs


andreachild:
The query which is being used by the utility to determine incoming and outgoing node labels from edges can be expensive for graphs with a lot of data. This query does not currently have a limit and can trigger out of memory issues on the graph provider.

```
let query = `MATCH (from)-[r:${sanitize(edgeStructure.label)}]->(to) RETURN DISTINCT labels(from) as fromLabel, labels(to) as toLabel`;
```

andreachild:
This has been resolved by setting a query limit using the existing `SAMPLE` constant that is used for other queries executed for schema discovery.

PR #87, opened 2025-05-02
Set limit on the expensive query which is retrieving edge directions

andreachild:
Set limit on the expensive query which is retrieving edge directions, otherwise large graphs could result in out of memory errors.

https://github.com/aws/amazon-neptune-for-graphql/issues/86

### DIFF
--- a/src/NeptuneSchema.js
+++ b/src/NeptuneSchema.js
@@ -178,8 +178,9 @@ async function getEdgesNames() {
 
 
 async function findFromAndToLabels(edgeStructure) {
-    let query = `MATCH (from)-[r:${sanitize(edgeStructure.label)}]->(to) RETURN DISTINCT labels(from) as fromLabel, labels(to) as toLabel`;
-    let response = await queryNeptune(query);
+    const query = `MATCH (from)-[r:${sanitize(edgeStructure.label)}]->(to) WITH from, to LIMIT $sample RETURN DISTINCT labels(from) as fromLabel, labels(to) as toLabel`;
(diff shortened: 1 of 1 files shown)


src/NeptuneSchema.js
```diff
@@ -178,8 +178,9 @@ async function getEdgesNames() {
 
 
 async function findFromAndToLabels(edgeStructure) {
-    let query = `MATCH (from)-[r:${sanitize(edgeStructure.label)}]->(to) RETURN DISTINCT labels(from) as fromLabel, labels(to) as toLabel`;
-    let response = await queryNeptune(query);
+    const query = `MATCH (from)-[r:${sanitize(edgeStructure.label)}]->(to) WITH from, to LIMIT $sample RETURN DISTINCT labels(from) as fromLabel, labels(to) as toLabel`;
+    loggerDebug(`Retrieving incoming and outgoing labels for edge ${edgeStructure.label} with limit ${SAMPLE}`, {toConsole: true});
+    const response = await queryNeptune(query, {sample: SAMPLE});
```
Cole-Greer, line 183: Where does `SAMPLE` come from? Is it already defined elsewhere?

andreachild: Yes it's being used by existing queries to limit the results. This particular query should have referenced it as well.

xiazcy (approved):

src/NeptuneSchema.js
xiazcy: Actually one follow-up comment, `SAMPLE` is defined at 5000 default above, and I don't see any changes, do we still want to increase the default limit to 100,000 or is that set else where?

andreachild: Andrea mentioned he wanted to keep the sample size consistent with the other queries so I did not change the value or create a separate value.

PR #93, opened 2025-05-09
Added separate input types for create and update mutations

flora-jin:
The utility currently generates the same input types for queries as mutations.

This is not ideal as it would be expected that the required fields will differ for queries vs mutations. For example, id should not be mandatory when creating a node as it should be auto generated by the server. However id should be required when updating a node.

Generating distinct input types for queries vs create and update mutations will provide better flexibility to the user to modify the generated schema accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,3 +48,4 @@ This release contains new support for Apollo Server integration.
 * Separated graphQL schema from resolver template ([#79](https://github.com/aws/amazon-neptune-for-graphql/pull/79))
 * Added unit tests for resolver and moved resolver integration tests to be unit tests ([#83](https://github.com/aws/amazon-neptune-for-graphql/pull/83))
 * Set limit on the expensive query which is retrieving distinct to and from labels for edges ([#89](https://github.com/aws/amazon-neptune-for-graphql/pull/89))
+* Added distinct input types for create and update mutations ([#93](https://github.com/aws/amazon-neptune-for-graphql/pull/93))
--- a/src/graphdb.js
+++ b/src/graphdb.js
@@ -81,6 +81,21 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
     checkForDuplicateNames(gdbs);
 
     gdbs.nodeStructures.forEach(node => {
+        // map with key=node property name and value=formatted string containing property name, type and optional alias
+        const nodePropertiesMap = new Map(node.properties.map(property => {
+            if (property.name === 'id') {
+                return [property.name, `\tid: ID\n`];
+            }
+            const sanitizedLabel = cleanseLabel(property.name);
+            let alias = '';
+            if (property.name !== sanitizedLabel) {
+                alias = ` @alias(property: "${property.name}")`;
+            }
+            return [property.name, `\t${sanitizedLabel}: ${property.type}${alias}\n`];
+        }));
+
+        const formattedNodeProperties = Array.from(nodePropertiesMap.values()).join('');
+
         // node type
         let nodeCase = cleanseLabel(node.label);
         if (changeCase) {
@@ -95,20 +110,7 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
         }
         
         r += '\t_id: ID! @id\n';
-
-        node.properties.forEach(property => {
-            if (property.name === 'id') {
-                r+= `\tid: ID\n`;
-            }
-            else {
-                let propertyCase = cleanseLabel(property.name);
-                let alias = '';
-                if (property.name !== propertyCase) {
-                    alias = ` @alias(property: "${property.name}")`;
-                }
-                r+= `\t${propertyCase}: ${property.type}${alias}\n`;
-            }
-        });
+        r += formattedNodeProperties;
         
         let edgeTypes = [];
         gdbs.edgeStructures.forEach(edge => {            
@@ -160,12 +162,10 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
             });
         });
 
-        const nodePropertyNames = new Set(node.properties.map((p) => p.name));
-
         // Add edge types
         edgeTypes.forEach((edgeType) => {
             // resolve any collision with node properties with the same name by adding an underscore prefix
-            const aliasedEdgeType = nodePropertyNames.has(edgeType)
+            const aliasedEdgeType = nodePropertiesMap.has(edgeType)
                 ? `_${edgeType}`
                 : edgeType;
 
@@ -182,16 +182,22 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
         // input for the node type
         r += `input ${nodeCase}Input {\n`;
         r += '\t_id: ID @id\n';
-        node.properties.forEach(property => {
-            let propertyCase = cleanseLabel(property.name);
-            if (property.name !== propertyCase) {
-                r+= `\t${propertyCase}: ${property.type} @alias(property: "${property.name}")\n`;
-            }
-            else {
-                r+= `\t${property.name}: ${property.type}\n`;
-            }
-        });
+        r += formattedNodeProperties;
         r += '}\n\n';
+
+        if (addMutations) {
+            // Create input for mutations
+            r += `input ${nodeCase}CreateInput {\n`;
+            r += '\t_id: ID @id\n';
+            r += formattedNodeProperties;
+            r += '}\n\n';
+
+            // Update input for mutations
+            r += `input ${nodeCase}UpdateInput {\n`;
+            r += '\t_id: ID! @id\n';
+            r += formattedNodeProperties;
+            r += '}\n\n';
+        }
     })
 
     const nodeLabels = new Set(gdbs.nodeStructures.map((n) => n.label));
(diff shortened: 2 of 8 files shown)


src/schemaModelValidator.js
```diff
@@ -138,19 +138,55 @@ function addNode(def) {
     let name = def.name.value;
     const idField = getIdField(def);
 
-    // Create Input type
-    typesToAdd.push(`input ${name}Input {\n${print(getInputFields(def))}\n}`);    
+    // General Input type
+    typesToAdd.push(`input ${name}Input {\n${print(getInputFields(def))}\n}`);
+
+    if (mutationsToAdd.length > 0) {
```
andreachild, line 144: This `if` condition should be removed because lines 183 and 184 always add the `create` and `update` mutations so we should always add the `CreateInput` and `UpdateInput`. Otherwise the generated schema could be invalid if the mutations are created but not the corresponding input types.

src/schemaModelValidator.js
```diff
@@ -138,19 +138,55 @@ function addNode(def) {
...
+                }
+            }
+        }
+        typesToAdd.push(`input ${name}CreateInput {\n${print(createFields)}\n}`);
+
+        // Update Input type
+        const updateFields = [];
+        for (const field of def.fields) {
+            if (isScalar(nullable(field.type))) {
+                if (field.type.kind === 'NonNullType' && !(field.type.type.name.value === GraphQLID.name)) {
```
andreachild, line 165: Should use optional chaining and `!==` instead of `!(x===y)`:
```
field.type?.kind === 'NonNullType' && field.type?.type?.name?.value !== GraphQLID.name
```

src/graphdb.js
andreachild, line 1: Please add tests for these changes to `graphdb.test.js`.

src/schemaModelValidator.js
andreachild, line 1: Please add tests for these changes to `schemaModelValidator.test.js`.

src/graphdb.js
```diff
@@ -192,6 +192,38 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
             }
         });
         r += '}\n\n';
+
+        if (addMutations) {
+            // Create input for mutations
+            r += `input ${nodeCase}CreateInput {\n`;
+            r += '\t_id: ID @id\n';
+            node.properties.forEach(property => {
+                let propertyCase = cleanseLabel(property.name);
```
andreachild, line 201: This code is repeated three times in this function so we can extract it into a reusable local variable like:

```
// map with key=node property name and value=formatted string containing property name, type and optional alias
const nodePropertiesMap = new Map(node.properties.map(property => {
    if (property.name === 'id') {
        return [property.name, `\tid: ID\n`];
    }
    const sanitizedLabel = cleanseLabel(property.name);
    let alias = '';
    if (property.name !== sanitizedLabel) {
        alias = ` @alias(property: "${property.name}")`;
    }
    return [sanitizedLabel, `\t${sanitizedLabel}: ${property.type}${alias}\n`];
}));
const formattedNodeProperties = Array.from(nodePropertiesMap.values()).join('');
```

And then these can be replaced:

`const nodePropertyNames = new Set(node.properties.map((p) => p.name));`

replaced with `nodePropertiesMap`

and each call to `node.properties.forEach` can be replaced with `formattedNodeProperties` like:

```
// input for the node type
r += `input ${nodeCase}Input {\n`;
r += '\t_id: ID @id\n';
r += formattedNodeProperties;
r += '}\n\n';

if (addMutations) {
    // Create input for mutations
    r += `input ${nodeCase}CreateInput {\n`;
    r += '\t_id: ID @id\n';
    r += formattedNodeProperties;
    r += '}\n\n';

    // Update input for mutations
    r += `input ${nodeCase}UpdateInput {\n`;
    r += '\t_id: ID! @id\n';
    r += formattedNodeProperties;
    r += '}\n\n';
}
```

src/test/schemaModelValidator.test.js
```diff
@@ -49,6 +49,28 @@ describe('validatedSchemaModel', () => {
...
+                    def.name.value === `${typeName}CreateInput`
+            );
+
+            const updateInputType = model.definitions.find(
+                def =>
+                    def.kind === 'InputObjectTypeDefinition' &&
+                    def.name.value === `${typeName}UpdateInput`
+            );
+
+            expect(createInputType).toBeDefined();
```
andreachild, line 69: Can you also add validation that id should be required for update but not create input?

src/graphdb.js
```diff
@@ -160,12 +162,10 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
             });
         });
 
-        const nodePropertyNames = new Set(node.properties.map((p) => p.name));
-
         // Add edge types
         edgeTypes.forEach((edgeType) => {
             // resolve any collision with node properties with the same name by adding an underscore prefix
-            const aliasedEdgeType = nodePropertyNames.has(edgeType)
+            const aliasedEdgeType = nodePropertiesMap.has(cleanseLabel(edgeType))
```
andreachild, line 168: I think something is weird here with the original logic but that can be investigated as part of a separate ticket. In the meantime it's best to restore this part to original logic by changing the map key to `property.name` instead of `sanitizedLabel` and then remove the call to `cleanseLabel` here.

src/schemaModelValidator.js
```diff
@@ -140,19 +140,53 @@ function addNode(def) {
...
+    typesToAdd.push(`input ${name}Input {\n${print(getInputFields(def))}\n}`);
+
     // Create Input type
-    typesToAdd.push(`input ${name}Input {\n${print(getInputFields(def))}\n}`);    
+    const createFields = [];
+    for (const field of def.fields) {
+        if (isScalarOrEnum(nullable(field.type))) {
+            if (field.type?.kind === 'NonNullType' && field.type?.type?.name?.value === GraphQLID.name) {
+                // make ID nullable by unwrapping the NonNullType from the field
+                const idFieldCopy = JSON.parse(JSON.stringify(field));
```
kmcginnes, line 152: What is happening here? Is this your way of doing a clone?

andreachild: Yes it's copying except unwrapping the type. I thought I had tried the spread before but I could be mistaken. @flora-jin could you pls try @kmcginnes's suggestion?

flora-jin: using the spread syntax will perform a shallow copy; is there a reason you did the deep copy previously @andreachild ?

src/test/graphdb.test.js
```diff
@@ -7,57 +7,69 @@ test('node with same property and edge label should add underscore prefix', () =
...
     loggerInit('./src/test/output', false, 'info');
-    const actual = inferGraphQLSchema('./src/test/node-edge-same-label-neptune-schema.json');
-    const expected = loadGraphQLSchema('./src/test/node-edge-same-label.graphql')
+    const actual = inferGraphQLSchema('./src/test/node-edge-same-label-neptune-schema.json', false);
+    const expected = loadGraphQLSchema('./src/test/node-edge-same-label.graphql');
     expect(actual).toBe(expected);
 });
 
-function inferGraphQLSchema(neptuneSchemaFilePath) {
+function inferGraphQLSchema(neptuneSchemaFilePath, addMutations) {
```
kmcginnes, line 70: Nit: Passing boolean arguments to functions is generally considered a code smell. This is just a test helper, so it's not hugely important. But it serves as a reason to bring up this topic.

The reason is because a reader of the code below would have no idea what the `false` or `true` value will do.

```js
inferGraphQLSchema('./src/test/node-edge-same-label-neptune-schema.json', false);
```

How would I change it? I would pass an `options` type that is an object as the parameter. This will force the caller to supply the parameter name as well as the value.

```js
inferGraphQLSchema(
  './src/test/node-edge-same-label-neptune-schema.json', 
  { addMutations: false }
);
```

And in the function definition you could provide a default value so that not all callers need to provide the options, only the ones that set it to true:

```js
function inferGraphQLSchema(neptuneSchemaFilePath, options = { addMutations: false })
```

src/test/schemaModelValidator.test.js
```diff
@@ -58,10 +58,36 @@ describe('validatedSchemaModel', () => {
...
+                    def.name.value === `${typeName}CreateInput`
+            );
+
+            const updateInputType = model.definitions.find(
+                def =>
+                    def.kind === 'InputObjectTypeDefinition' &&
+                    def.name.value === `${typeName}UpdateInput`
+            );
+
+            const createIdField = getIdFields(createInputType)[0];
```
kmcginnes, line 77: These calls will fail if `createInputType` is undefined. I suggest moving them to below your expectation that they are defined so that tests will fail in a helpful way.

src/schemaModelValidator.js
andreachild: I can't recall when I was doing the POC for this fix. If it works as a shallow copy then it should be OK to change it to shallow. I would not think that the ID field would have deep nested data.

CHANGELOG.md
```diff
@@ -44,3 +44,4 @@ This release contains new support for Apollo Server integration.
 * Separated graphQL schema from resolver template (#79)
 * Added unit tests for resolver and moved resolver integration tests to be unit tests (#83)
 * Set limit on the expensive query which is retrieving distinct to and from labels for edges (#89)
+* Added distinct input types for create and update mutations (#93)
```
kmcginnes, line 47: Should this be the newer style of link I saw in @andreachild's other PR?

```
[#93](https://github.com/aws/amazon-neptune-for-graphql/pull/93)
```

src/schemaModelValidator.js
kmcginnes: Even if it did, as long as you don't modify any of that data later you should be good just copying it with the spread.

The only time you'll get in to trouble is when mutating the data.

CHANGELOG.md
flora-jin: @andreachild I checked the changelog changes for the most recent enum change and the formatting seems to still just only include the number of the pull request. Which newer style is Kris referring to ?

andreachild: This one I just merged: https://github.com/aws/amazon-neptune-for-graphql/pull/97/files

kmcginnes (approved):

flora-jin:
@kmcginnes Still haven't pushed my changes for the comments to the PR yet 😅

kmcginnes (approved):

PR #94, opened 2025-05-09
Fix failing integration test by excluding node_modules from apollo package

andreachild:
Fix failing integration test by excluding node_modules from apollo zip archive, as it is not needed since users are expected to run npm install. Lambda packages still require node_modules in the zip archive. Also added missing changelog entries from previous commits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ This release contains new support for Apollo Server integration.
 * Fix query from AppSync with an empty filter object (#61)
 * Retain numeric parameter value type when creating open cypher query (#63)
 * Fixed bug with ID argument type conversion and added Apollo arguments to help menu (#74)
+* Upgraded axios and babel versions to fix security warnings (#90)
+* Fixed failing integration test by excluding `node_modules` from Apollo zip (#94)
 
 ### Features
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -40,3 +42,4 @@ This release contains new support for Apollo Server integration.
 * Add template file missing from packaging (#71)
 * Separated graphQL schema from resolver template (#79)
 * Added unit tests for resolver and moved resolver integration tests to be unit tests (#83)
+* Set limit on the expensive query which is retrieving distinct to and from labels for edges (#89)
```
kmcginnes, line 45: Are these extra changelog entries intentional?

kmcginnes (commented):
Left a couple of questions/suggestions

CHANGELOG.md
andreachild: Yes I had forgotten to include them as part of https://github.com/aws/amazon-neptune-for-graphql/pull/89

src/zipPackage.js
```diff
@@ -95,7 +102,9 @@ export async function createApolloDeploymentPackage({zipFilePath, resolverFilePa
             // querying neptune using SDK not yet supported
             {source: path.join(modulePath, '/../templates/queryHttpNeptune.mjs')}
         ],
-        includeContent: [{source: envVars.join('\n'), target: '.env'}]
+        includeContent: [{source: envVars.join('\n'), target: '.env'}],
+        // exclude node_modules from apollo package
+        excludePatterns: ['node_modules/**', '**/node_modules/**']
```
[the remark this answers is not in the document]
andreachild: Good point. I tested it out and found I only needed `**/node_modules/**` to exclude both root and nested so I removed the other pattern which was unnecessary. I tried just `node_modules` but that wasn't sufficient.

kmcginnes (approved):

PR #95, opened 2025-05-15
Allow enum types in schema

andreachild:
Fix issues with schema generation which was causing input schema enum types to be generated as edge types and prevented enum fields from being used as part of input types.

Example of generated schema before fix:
```
type Todo {
  _id: ID!
  name: String
  description: String
  category: Category
  # category enum should not produce an edge field
  categoryEdge: CategoryEdge
}
enum Category {
  CAT1
  CAT2
}
# category enum should not produce an edge type
type CategoryEdge {
  _id: ID!
}
input TodoInput {
  _id: ID!
  name: String
  description: String
  # category enum field missing from input
}
```

After fix:
```
type Todo {
  _id: ID!
  name: String
  description: String
  category: Category
}
enum Category {
  CAT1
  CAT2
}
input TodoInput {
  _id: ID!
  name: String
  description: String
  category: Category
}
```

### DIFF
--- a/src/schemaModelValidator.js
+++ b/src/schemaModelValidator.js
@@ -16,9 +16,11 @@ import {gql} from 'graphql-tag'
 import { loggerInfo, yellow } from "./logger.js";
 
 let quiet = false;
+// TODO change variables to local scope instead of global so this module can be used against multiple schemas
 const typesToAdd = [];
 const queriesToAdd = [];
 const mutationsToAdd = [];
+const enumTypes = [];
 
 function lowercaseFirstCharacter(inputString) {
     if (inputString.length === 0) {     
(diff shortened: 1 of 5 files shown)


src/schemaModelValidator.js
```diff
@@ -16,9 +16,11 @@ import {gql} from 'graphql-tag'
 import { loggerInfo, yellow } from "./logger.js";
 
 let quiet = false;
+// TODO change variables to local scope instead of global so this module can be used against multiple schemas
```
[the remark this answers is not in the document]
andreachild: No, was going to leave that for another time

templates/queryHttpNeptune.mjs
```diff
@@ -33,6 +33,9 @@ export async function queryNeptune(neptuneUrl, resolvedQuery, options = {logging
         };
 
         let response;
+        if (loggingEnabled) {
+            console.log("Query: ", JSON.stringify(resolvedQuery, null, 2));
+        }
```
kmcginnes, line 38: This looks like you intentionally left it in, but I wanted to ask just to make sure it wasn't just for debugging.

kmcginnes (approved):

templates/queryHttpNeptune.mjs
andreachild: Yes I did this intentionally because I got tired of adding it each time I needed to work on the code.

PR #96, opened 2025-05-15
Account for id fields which may be missing the @id directive

flora-jin:
Currently, if an input schema does not contain the @id directive for id fields, the utility does not recognize it as an id field and adds another id field with the @id directive. 

PR for fixing duplication bug.


andreachild:
LGTM

kmcginnes (approved):

PR #97, opened 2025-05-15
Allow custom scalars in schema

andreachild:
If a given input schema contained a custom scalar, the utility would generate an invalid schema and any field which has a custom scalar type will not be able to be used as input for a query or mutation.

Changes include:
* fixed `schemaModelValidator.js` to no longer interpret custom scalars as edges and also allow fields of custom scalar types to be included as part of input types
* fixed `schemaModelValidator.js` to add the `Options` parameter to generated queries that retrieve an array of objects instead of the queries that retrieve single objects
* added documentation for how to use custom scalars with Apollo Server and scalar restrictions with App Sync

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,28 +20,30 @@ This release contains new support for Apollo Server integration.
 
 ### Bug Fixes
 
-* Don't cast integers to floats in Neptune schema (#62)
-* Fix query from AppSync with an empty filter object (#61)
-* Retain numeric parameter value type when creating open cypher query (#63)
-* Fixed bug with ID argument type conversion and added Apollo arguments to help menu (#74)
-* Upgraded axios and babel versions to fix security warnings (#90)
-* Fixed failing integration test by excluding `node_modules` from Apollo zip (#94)
-* Fixed enum types in schema to be included in input types (#95)
-* Fixed bug where id fields without @id directives are not accounted for (#96)
+* Don't cast integers to floats in Neptune schema ([#62](https://github.com/aws/amazon-neptune-for-graphql/pull/62))
+* Fix query from AppSync with an empty filter object ([#61](https://github.com/aws/amazon-neptune-for-graphql/pull/61))
+* Retain numeric parameter value type when creating open cypher query ([#63](https://github.com/aws/amazon-neptune-for-graphql/pull/63))
+* Fixed bug with ID argument type conversion and added Apollo arguments to help menu ([#74](https://github.com/aws/amazon-neptune-for-graphql/pull/74))
+* Upgraded axios and babel versions to fix security warnings ([#90](https://github.com/aws/amazon-neptune-for-graphql/pull/90))
+* Fixed failing integration test by excluding `node_modules` from Apollo zip ([#94](https://github.com/aws/amazon-neptune-for-graphql/pull/94))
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -20,28 +20,30 @@ This release contains new support for Apollo Server integration.
...
+* Fix query from AppSync with an empty filter object ([#61](https://github.com/aws/amazon-neptune-for-graphql/pull/61))
+* Retain numeric parameter value type when creating open cypher query ([#63](https://github.com/aws/amazon-neptune-for-graphql/pull/63))
+* Fixed bug with ID argument type conversion and added Apollo arguments to help menu ([#74](https://github.com/aws/amazon-neptune-for-graphql/pull/74))
+* Upgraded axios and babel versions to fix security warnings ([#90](https://github.com/aws/amazon-neptune-for-graphql/pull/90))
+* Fixed failing integration test by excluding `node_modules` from Apollo zip ([#94](https://github.com/aws/amazon-neptune-for-graphql/pull/94))
+* Fixed enum types in schema to be included in input types ([#95](https://github.com/aws/amazon-neptune-for-graphql/pull/95))
+* Fixed bug where id fields without @id directives are not accounted for ([#96](https://github.com/aws/amazon-neptune-for-graphql/pull/96))
+* Fixed custom scalar types in schema to be included in input types and queries generated from an input schema which retrieve an array to have an option parameter with limit ([#95](https://github.com/aws/amazon-neptune-for-graphql/pull/97))
```
kmcginnes, line 31: Nit: It might make sense to break this in to two entries. They'll both point to the same PR, but I think that is ok for the clarity of the changelog.

kmcginnes: Also, the title of the link is wrong. Should be `[#97]`

```
[#95](https://github.com/aws/amazon-neptune-for-graphql/pull/97)
```

src/schemaModelValidator.js
```diff
@@ -294,6 +298,7 @@ function inferGraphDatabaseDirectives(schemaModel) {
                 // add relationships
                 def.fields.forEach(field => {                    
                     if (field.type.type !== undefined) {
+                        // FIXME handle NonNullType wrapper
```
kmcginnes, line 301: Is this being saved for a future PR?

src/test/schemaModelValidator.test.js
```diff
@@ -1,15 +1,15 @@
 import { readFileSync } from 'node:fs';
 import { loggerInit } from '../logger.js';
 import { validatedSchemaModel } from '../schemaModelValidator.js';
-import { schemaParser } from '../schemaParser.js';
+import {schemaParser, schemaStringify} from '../schemaParser.js';
```
kmcginnes, line 4: Nit: It looks like you lost the formatting here (the spaces inside the curly braces)

kmcginnes (commented):
Nice changes overall.

src/schemaModelValidator.js
andreachild: Yes I am going to make a separate ticket for it, this was just to remind me when I'm dealing with this code later

CHANGELOG.md
andreachild: Thanks, changed as suggested

src/test/schemaModelValidator.test.js
andreachild: Fixed formatting, I'll need to change my IDE style to be consistent

kmcginnes (approved):

PR #98, opened 2025-05-21
Enabled mutations for apollo server

flora-jin:
As mutations are currently unsupported in the apollo server, they have been enabled and verified that they work as expected when executed against the graphQL API.

### DIFF
--- a/templates/ApolloServer/index.mjs
+++ b/templates/ApolloServer/index.mjs
@@ -14,19 +14,47 @@ const queryDefinition = typeDefs.definitions.find(
 );
 const queryNames = queryDefinition ? queryDefinition.fields.map(field => field.name.value) : [];
 
+const mutationDefinition = typeDefs.definitions.find(
+    definition => definition.kind === 'ObjectTypeDefinition' && definition.name.value === 'Mutation'
+);
+const mutationNames = mutationDefinition ? mutationDefinition.fields.map(field => field.name.value) : [];
+
+/**
+ * Resolves GraphQL queries and mutations by processing the info object and arguments.
+ *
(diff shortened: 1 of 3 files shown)


andreachild:
Somewhere in the documentation is a statement about how mutations are not yet supported for Apollo - that should be updated with this change.

templates/ApolloServer/index.mjs
```diff
@@ -30,6 +34,21 @@ const resolvers = {
         };
         return accumulator;
     }, {}),
+
+    Mutation: mutationNames.reduce((accumulator, mutationName) => {
+        accumulator[mutationName] = (parent, args, context, info) => {
+            const event = {
+                field: info.fieldName,
+                arguments: args,
+                selectionSet: info.fieldNodes[0].selectionSet,
```
kmcginnes, line 43: Are we guaranteed to have at least one value in `info.fieldNodes` array? If not, then we should check for `info.fieldNodes.length > 0` before indexing.

flora-jin: @kmcginnes What about using optional chaining for the check such as 
```
info.fieldNodes?.[0]?.selectionSet ?? null
```

kmcginnes: That certainly will keep you from having a runtime error. But we should have an assumption of what to expect. Perhaps an error is appropriate if the `fieldNodes` array is empty 🤷 . That's why I asked.

Also, assigning `null` to `selectionSet` might not be desired, depending on the logic after. So, as always, it depends. 🤓

andreachild (approved):

PR #99, opened 2025-05-24
Integration test improvements

andreachild:
Integration test improvements:
* fixed integration tests that were broken due to differing results returned from neptune when the query limit is less than the number of potential results by changing the queries such that they will have a known set of results
* fixed neptune query result verification that was failing for expected results that contained an array
* added test output folder cleanup
* removed tests that checked resolver file size and content as they are not useful now that the schema was separated from the resolver (and these tests require updating any time the resolver is changed)
* replaced resolver file size and content checks with checks that test output folder contains expected file
* removed integration tests which check the resolved open cypher query as graphQL->cypher query resolution is now covered by the resolver unit test


test/TestCases/Case04/Case04.02.test.js
```diff
@@ -10,9 +9,17 @@ checkOutputFilesSize('./test/TestCases/Case04/output', casetest.testOutputFilesS
...
 const refNeptuneSchema = readJSONFile('./test/TestCases/Case04/outputReference/output.neptune.schema.json');
 
-checkOutputFileContent(
+describe('Validate neptune schema', () => {
+    afterAll(async () => {
+        fs.rmSync('./test/TestCases/Case04/output', {recursive: true});
+    });
+
+    // note that this test can be flaky depending on how the air routes sample data was loaded into neptune
+    // for more consistent results, use neptune notebook %seed with gremlin language
```
Cole-Greer, line 18: Just curious, what ends up being flaky here, does the content of the schema differ based on load order (due to limited sampling) or is the content of the schema the same, but somehow ordered or formatted differently?

Cole-Greer (approved):
LGTM

test/TestCases/Case04/Case04.02.test.js
andreachild: I noticed that the neptune data itself can be different depending on if the sample data was loaded via notebook `%seed` magic using cypher vs gremlin. Specifically the `id` of nodes and edges. When I loaded the sample data using `cypher`, this test would fail. When I loaded the sample data with `gremlin` it would pass.

PR #100, opened 2025-05-26
Enhanced string filtering

andreachild:
This changeset modifies the `filter` query argument to support string comparison operators `eq`, `contains`, `startsWith`, and `endsWith`:
* added `StringScalarFilters` type which replaces `String` fields for query input types (mutation input types not affected)
* refactored the resolver to extract filters and arguments as objects from a given graphQL query instead of using string concatenation to increase code reusability
* added logic to resolver to convert filter string comparison operators to cypher equivalent syntax
* fixed usage of variables for nested edge subqueries by passing along query variables as part of apollo and app sync event handling - this is required to support the usage of string comparison operators via variables on the subqueries
* fixed nested edge subqueries to return an empty array if no results were produced using `CASE` - previously null values would be returned, producing an invalid graphQL response
* changed `graphdb.test.js` to use `prettier` graphQL formatter to account for whitespace differences between expected and actual schemas

### DIFF
--- a/src/graphdb.js
+++ b/src/graphdb.js
@@ -74,6 +74,57 @@ function cleanseLabel(label) {
         .replaceAll("-", "_hy_");
 }
 
+/**
+ * Extracts node properties, ensuring each node has an ID property and any properties with special characters are 
+ * sanitized and aliased accordingly.
+ *
+ * @param {Object} node - The node object containing properties
+ * @param {Array<Object>} node.properties - Array of property objects
+ * @param {string} node.properties[].name - Property name
+ * @param {string} node.properties[].type - Property type
+ * @returns {Array<Object>} Formatted property objects with name, optional alias, and type
+ */
+function extractNodeProperties(node= {}) {
+    if (!node?.properties) {
+        return [];
+    }
+    const nodeProperties = node.properties.map(property => {
+        const sanitizedLabel = cleanseLabel(property.name);
+        return {
+            name: sanitizedLabel,
+            alias: property.name !== sanitizedLabel ? property.name : null,
+            type: property.name === 'id' ? 'ID' : property.type
+        };
+    });
+    if (!nodeProperties.some(property => property.type === 'ID')) {
+        nodeProperties.unshift({
+            name: '_id',
+            alias: null,
+            type: 'ID'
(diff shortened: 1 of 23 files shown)


src/graphdb.js
```diff
@@ -74,28 +74,64 @@ function cleanseLabel(label) {
...
+ * @param {Array<Object>} node.properties - Array of property objects
+ * @param {string} node.properties[].name - Property name
+ * @param {string} node.properties[].type - Property type
+ * @returns {Array<Object>} Formatted property objects with name, optional alias, and type
+ */
+function extractNodeProperties(node= {}) {
+    if (!node?.properties) {
+        return [];
+    }
+    const nodeProperties = node.properties.map(property => {
```
Cole-Greer, line 91: Is it guaranteed that `node.properties` is an array?

andreachild: Yes the utility generates a json schema of the neptune db and instantiates the each `node.properties` as an empty array. It then adds properties to the array after executing queries to discover the properties for the node type (this logic is in `NeptuneSchema.js`.

src/test/templates/JSResolverOCHTTPS.test.js
```diff
@@ -13,22 +13,22 @@ schemaDataModel = JSON.stringify(schemaDataModel, null, 2);
 const schemaModel = JSON.parse(schemaDataModel);
 initSchema(schemaModel);
 
-test('should resolve queries with a filter', () => {
+test('should resolve app sync event queries with a filter', () => {
     const result = resolveGraphDBQueryFromAppSyncEvent({
         field: 'getNodeAirport',
-        arguments: { filter: { code: 'SEA' } },
+        arguments: { filter: { code: {eq: 'SEA'} } },
```
Cole-Greer, line 19: Is the new `eq` syntax here mandatory? Or is the old filter syntax still supported?

andreachild: The `eq` is now mandatory, the old filtering syntax will no longer be supported. There's no way to specify in the graphQL schema that the `code ` can be a `string` (old syntax) OR a `StringScalarFilter` (new syntax).

Issue #84, opened 2025-04-24
Support for Fragments


andreachild:
The graphQL utility does not currently support execution of queries that use fragments as shown in the screenshots below. Any fields in fragments will have null values when the query is executed.

![Image](https://github.com/user-attachments/assets/698a5408-6f20-4676-b7ac-f9bc3434658b)

![Image](https://github.com/user-attachments/assets/d2f0edae-1dcf-43e7-9101-97e645aab76f)

andreachild:
Apollo fragment support was done by https://github.com/aws/amazon-neptune-for-graphql/pull/103

PR #103, opened 2025-06-04
Support for query fragments with Apollo

andreachild:
https://github.com/aws/amazon-neptune-for-graphql/issues/84

This changeset enables the use of graphQL query fragments with Apollo Server:
* changed Apollo event handler to pass along any query fragments to the resolver
* added logic to the resolver to detect referenced fragments in the query AST and replace them with the appropriate fragment's selection set
* changed App Sync event handlers to throw an Error if a fragment is detected in the query so that the user is presented with an appropriate error message 'Fragments are not supported' instead of failing downstream and presenting a cryptic error message
* changed functions in the resolver which require fragments data to accept a single object argument instead of multiple arguments for easier readability when calling the function
* updated `airports.customized.graphql` test schema to have separate input types for create and update mutations so that the resolver unit test could verify test cases for mutations with fragments
* fixed a couple resolver unit tests that were not correctly wrapping the function call when using `expect().toThrow(msg)`
* note that this changeset does not address inline fragments

### DIFF
--- a/src/pipelineResources.js
+++ b/src/pipelineResources.js
@@ -579,9 +579,9 @@ export function request(ctx) {
     
 export function response(ctx) {
     const { error, result } = ctx;
-      if (error) {
+    if (error) {
         util.appendError(error.message, error.type, result);
-      }
+    }
     return result;
 }`
 
--- a/src/pipelineResources.js
+++ b/src/pipelineResources.js
@@ -578,7 +578,11 @@ export function request(ctx) {
 }
     
 export function response(ctx) {
-    return ctx.result;
+    const { error, result } = ctx;
+      if (error) {
+        util.appendError(error.message, error.type, result);
+      }
+    return result;
 }`
 
     };
--- a/templates/JSResolverOCHTTPS.js
+++ b/templates/JSResolverOCHTTPS.js
@@ -40,6 +40,8 @@ export function resolveGraphDBQueryFromAppSyncEvent(event) {
     return resolveGraphDBQueryFromEvent({
         field: event.field,
         arguments: event.arguments,
+        // fragments not yet supported in app sync - see https://github.com/aws-samples/appsync-with-postgraphile-rds/issues/18
+        fragments: {},
         variables: event.variables,
         selectionSet: event.selectionSetGraphQL ? gql`${event.selectionSetGraphQL}`.definitions[0].selectionSet : {}
     });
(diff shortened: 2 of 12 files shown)


src/pipelineResources.js
```diff
@@ -578,7 +578,11 @@ export function request(ctx) {
 }
     
 export function response(ctx) {
-    return ctx.result;
+    const { error, result } = ctx;
+      if (error) {
+        util.appendError(error.message, error.type, result);
+      }
+    return result;
```
kmcginnes, line 585: Nit: It seems like the indentation is off here.

templates/JSResolverOCHTTPS.js
```diff
@@ -886,14 +921,47 @@ function convertToValueNode(value) {
...
+        return;
+    }
+
+    // Find all fragment spreads referenced in query selection
+    const fragmentSpreads = selection.selectionSet.selections.reduce((acc, fragmentSelection, index) => {
+        if (fragmentSelection?.kind === 'FragmentSpread' && fragmentSelection?.name?.value) {
+            if (fragments[fragmentSelection.name.value]) {
+                acc.push({ fragmentSelection, index });
+            } else {
+                throw new GraphQLError(`Fragment ${fragmentSelection.name.value} not found`);
```
kmcginnes, line 949: Just double checking that we want the entire process to error if a fragment is not found.

Alternatively, we could process missing fragments as we do today, with null values.

andreachild: I think an error in this case is appropriate. This means the query that was sent was invalid as it references fragments that were not included in the query. This is actually an edge case and not possible to trigger via Apollo/App Sync as there is validation that all fragments referenced in the query exist.

templates/JSResolverOCHTTPS.js
```diff
@@ -993,12 +1077,20 @@ function resolveGrapgDBqueryForGraphQLMutation (obj, querySchemaInfo) {
...
             return `${nodeName}.${arg.name} = $${param}`;
         }).join(', ');
+        // FIXME handle update mutations with selection set that contains an edge
+        // example:
+        // updateNodeAirport(input: $input) {
+        //     id
+        //     airportRoutesIn {
+        //       code
+        //     }
+        //   }
```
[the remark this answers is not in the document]
andreachild: Yes I tend to put FIXME comments whenever I find bugs unrelated to what I'm working on so I don't forget about it later.

PR #105, opened 2025-06-07
Added support for queries with sorting

flora-jin:
The current graphQL utility does not support the sort argument in queries.

Added support for sort argument in queries including nested.
Wrote unit tests and fixed previous unit tests to follow added support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ This release contains new support for Apollo Server integration.
 * Support output of zip package of Apollo Server artifacts (([#70](https://github.com/aws/amazon-neptune-for-graphql/pull/70)), ([#72](https://github.com/aws/amazon-neptune-for-graphql/pull/72)), ([#73](https://github.com/aws/amazon-neptune-for-graphql/pull/73)), ([#75](https://github.com/aws/amazon-neptune-for-graphql/pull/75)), ([#76](https://github.com/aws/amazon-neptune-for-graphql/pull/76)))
 * Allow filtering using string comparison operators `eq`, `contains`, `startsWith`, `endsWith` (([#100](https://github.com/aws/amazon-neptune-for-graphql/pull/100))
 * Added pagination support through the addition of an `offset` argument in query options which can be used in combination with the existing `limit` (([#102](https://github.com/aws/amazon-neptune-for-graphql/pull/102))
+* Added support for queries with sorting ([#105](https://github.com/aws/amazon-neptune-for-graphql/pull/105))
 
 
 ### Improvements
--- a/src/graphdb.js
+++ b/src/graphdb.js
@@ -125,11 +125,17 @@ function formatProperties(properties = [], typeOverrides = new Map()) {
     }).join('');
 }
 
-function graphDBInferenceSchema (graphbSchema, addMutations) {
+function graphDBInferenceSchema (graphdbSchema, addMutations) {
     let r = '';
-    const gdbs = JSON.parse(graphbSchema);
+    const gdbs = JSON.parse(graphdbSchema);
 
     checkForDuplicateNames(gdbs);
+
+    // sorting direction enum
+    r += `enum SortingDirection {\n`;
+    r += '\tASC\n';
+    r += '\tDESC\n';
+    r += '}\n\n';
 
     gdbs.nodeStructures.forEach(node => {
         // node type
@@ -157,8 +163,8 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
 
                 if (direction.from === node.label && direction.to === node.label){
                     if (direction.relationship === 'MANY-MANY') {
-                        r += `\t${nodeCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${nodeCase}Input, options: Options): [${nodeCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
-                        r += `\t${nodeCase.toLocaleLowerCase() + edgeCase}sIn(filter: ${nodeCase}Input, options: Options): [${nodeCase}] @relationship(edgeType:"${edge.label}", direction:IN)\n`;
+                        r += `\t${nodeCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${nodeCase}Input, options: Options, sort: [${nodeCase}Sort!]): [${nodeCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
+                        r += `\t${nodeCase.toLocaleLowerCase() + edgeCase}sIn(filter: ${nodeCase}Input, options: Options, sort: [${nodeCase}Sort!]): [${nodeCase}] @relationship(edgeType:"${edge.label}", direction:IN)\n`;
                     }
                     if (direction.relationship === 'ONE-ONE') {
                         r += `\t${nodeCase.toLocaleLowerCase() + edgeCase}Out: ${nodeCase} @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
@@ -170,10 +176,10 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
                 
                 if (direction.from === node.label && direction.to !== node.label){
                     if (direction.relationship === 'MANY-MANY') {
-                        r += `\t${toCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${toCase}Input, options: Options): [${toCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
+                        r += `\t${toCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${toCase}Input, options: Options, sort: [${toCase}Sort!]): [${toCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
                     }
                     if (direction.relationship === 'ONE-MANY') {
-                        r += `\t${toCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${toCase}Input, options: Options): [${toCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
+                        r += `\t${toCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${toCase}Input, options: Options, sort: [${toCase}Sort!]): [${toCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
                     }
                     if (direction.relationship === 'MANY-ONE') {
                         r += `\t${toCase.toLocaleLowerCase() + edgeCase}Out: ${toCase} @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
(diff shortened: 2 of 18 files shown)


kmcginnes (approved):
Code looks great

CHANGELOG.md
```diff
@@ -56,4 +56,5 @@ This release contains new support for Apollo Server integration.
 * Added distinct input types for create and update mutations ([#93](https://github.com/aws/amazon-neptune-for-graphql/pull/93))
 * Enabled mutations for the Apollo Server ([#98](https://github.com/aws/amazon-neptune-for-graphql/pull/98))
 * Refactored integration tests to be less vulnerable to resolver logic changes ([#99](https://github.com/aws/amazon-neptune-for-graphql/pull/99))
-* Enabled usage of query fragments with Apollo Server ([#103](https://github.com/aws/amazon-neptune-for-graphql/pull/103))
\ No newline at end of file
+* Enabled usage of query fragments with Apollo Server ([#103](https://github.com/aws/amazon-neptune-for-graphql/pull/103))
+* Added support for queries with sorting ([#105](https://github.com/aws/amazon-neptune-for-graphql/pull/105))
```
andreachild, line 60: I think this can go under features!

src/graphdb.js
```diff
@@ -125,12 +125,18 @@ function formatProperties(properties = [], typeOverrides = new Map()) {
...
-function graphDBInferenceSchema (graphbSchema, addMutations) {
+function graphDBInferenceSchema (graphdbSchema, addMutations) {
     let r = '';
-    const gdbs = JSON.parse(graphbSchema);
+    const gdbs = JSON.parse(graphdbSchema);
 
     checkForDuplicateNames(gdbs);
 
+    // sorting direction enum
+    r += `enum SortingDirection {\n`;
```
andreachild, line 135: Nit: can this be moved to the location where other static types are added? ie. above the `Options` and `StringScalarFilters` input types.

src/graphdb.js
```diff
@@ -231,6 +237,20 @@ function graphDBInferenceSchema (graphbSchema, addMutations) {
             r += formatProperties(nodeProperties, new Map([['ID', 'ID! @id']]));
             r += '}\n\n';
         }
+
+        // sort input
+        r += `input ${nodeCase}Sort {\n`;
+        r += '\t_id: SortingDirection\n';
```
andreachild, line 243: It's possible that the `ID` field is not named `_id`. It would be better to reuse the `formatProperties` function that accounts for ID fields that have different names. You can do this:

```
// sort input
const sortProperties = nodeProperties.map(prop => ({
    ...prop,
    type: 'SortingDirection'
}));
r += `input ${nodeCase}Sort {\n`;
r += formatProperties(sortProperties);
r += '}\n\n';
```

src/schemaModelValidator.js
```diff
@@ -306,6 +339,9 @@ function inferGraphDatabaseDirectives(schemaModel) {
         .filter(definition => definition.kind === 'EnumTypeDefinition')
         .forEach(definition => enumTypes.push(definition.name.value));
 
+    // Generate sort enum
+    typesToAdd.push(`enum SortDirection {\nASC\nDESC\n}`);
```
andreachild, line 343: Can this be moved to where other static types are declared? ie. `Options` and `StringScalarFilters`

src/test/templates/JSResolverOCHTTPS.test.js
```diff
@@ -200,6 +218,51 @@ test('should resolve app sync event with nested edge filters and variables', ()
     });
 });
 
+test('should resolve app sync event with sort arguments as a list', () => {
```
andreachild, line 221: Is this a duplicate of 'should resolve queries with sort arguments as a list'?

templates/JSResolverOCHTTPS.js
```diff
@@ -482,6 +485,19 @@ function extractPositiveIntegerFieldValue(field) {
     throw new GraphQLError(`The ${field.name.value} value must be a positive integer`);
 }
 
+function getSortInfo(fields, schemaInfo) {
```
andreachild, line 488: This should be named something like `setSortInfo` because it's setting values on the `schemaInfo` and not returning values.

templates/JSResolverOCHTTPS.js
```diff
@@ -482,6 +485,19 @@ function extractPositiveIntegerFieldValue(field) {
...
+    if (Object.keys(fields)?.length > 1) {
```
andreachild, line 489: This check is essentially throwing an error if the given array of fields has more than one item, therefore the `fields.forEach` which is called afterwards can just be changed to be called on a single field instead because at that point there is either no field or only one field.

templates/JSResolverOCHTTPS.js
```diff
@@ -482,6 +485,19 @@ function extractPositiveIntegerFieldValue(field) {
...
+        // too many keys
+        throw new Error('Cannot have more than one key in sort object');
+    }
+
+    fields.forEach(field => {
+        const fieldName = field.name.value;
+        const direction = field.value.value.toUpperCase();
+        schemaInfo.argOrderBy.push(`${fieldName} ${direction}`);
```
andreachild, line 497: With the resolver I would like to slowly move towards all query string formatting done at the end of the logic, which makes it easier to maintain. Could this be changed to set a list of objects for `argOrderBy` instead of a list of strings? And then the code that is referencing `argOrderBy` can do the string concatenation.
example:

```
schemaInfo.argOrderBy.push({field: fieldName, direction: direction});
... 
const sortClause = querySchemaInfo.argOrderBy.length > 0 ? ` ORDER BY ${querySchemaInfo.argOrderBy.map(orderBy => `${querySchemaInfo.pathName}.${orderBy.field} ${orderBy.direction}`).join(', ')}` : '';
```

templates/JSResolverOCHTTPS.js
```diff
@@ -971,7 +997,7 @@ function replaceFragmentSelections(selection, fragments) {
  */
 function extractCypherFieldsFromArgumentFields(queryArgumentFields, schemaInfo) {
     return queryArgumentFields.reduce((cypherFields, field) => {
-        const matchingArg = schemaInfo.args?.find(arg => field.name?.value === arg.name)
+        const matchingArg = schemaInfo.args?.find(arg => field.name?.value === arg.name && arg.type !== 'SortingDirection');
```
andreachild, line 1000: What kind of query requires this change? When I revert this change all of the tests still pass.

src/graphdb.js
flora-jin: Moving this would cause the order of the enum placement to change. I believed that as all the input types will include a sort type which references the enum the enum definition should be placed always at the beginning of the schema ?

src/schemaModelValidator.js
flora-jin: If this is moved that would change the order of the enum definition placement. Don't think the order matters for the schema but was wondering of the general formatting as I thought enum definition would go before input type declaration.

src/graphdb.js
andreachild: The order of types in the schema does not seem to matter, looking at the existing test schemas. But that's OK it can stay where it is so you don't have to update all of the test files.

templates/JSResolverOCHTTPS.js
andreachild: It was probably addressed after I had changed some looping logic to use filter instead for better efficiency.

templates/JSResolverOCHTTPS.js
flora-jin: Still need to update this part btw

templates/JSResolverOCHTTPS.js
```diff
@@ -688,6 +712,10 @@ function createTypeFieldStatementAndRecurse({selection, fieldSchemaInfo, lastNam
     const thisWithId = withStatements.push({carryOver: schemaTypeInfo.pathName, inLevel: '', content: ''}) - 1;
 
     if (schemaTypeInfo.isArray) {
+        const orderByClause = fieldSchemaInfo.argOrderBy?.length ? `WITH ${lastNamePath}, ${schemaTypeInfo.pathName} ORDER BY ${fieldSchemaInfo.argOrderBy.map(orderBy => `${schemaTypeInfo.pathName}.${orderBy.field} ${orderBy.direction}`).join(', ')}` : '';
+        if (orderByClause) {
+            matchStatements.push(orderByClause);
```
andreachild, line 717: Can this be moved to where the match statement is built above? 
ie:

```
const orderByClause = fieldSchemaInfo.argOrderBy?.length ? ` WITH ${lastNamePath}, ${schemaTypeInfo.pathName} ORDER BY ${fieldSchemaInfo.argOrderBy.map(orderBy => `${schemaTypeInfo.pathName}.${orderBy.field} ${orderBy.direction}`).join(', ')}` : '';

if (schemaTypeInfo.isRelationship) {
    const arrows = ['<-', '-', '->'];
    if (schemaTypeInfo.relationship.direction !== 'IN') {
        arrows.shift();
    }
    matchStatements.push(`OPTIONAL MATCH (${lastNamePath})${arrows[0]}[${schemaTypeInfo.pathName}_${schemaTypeInfo.relationship.edgeType}:${schemaTypeInfo.relationship.edgeType}]${arrows[1]}(${schemaTypeInfo.pathName}:\`${schemaTypeInfo.typeAlias}\`${queryArgs})${whereClause}${orderByClause}`);
}
```

templates/JSResolverOCHTTPS.js
```diff
@@ -482,6 +485,20 @@ function extractPositiveIntegerFieldValue(field) {
     throw new GraphQLError(`The ${field.name.value} value must be a positive integer`);
 }
 
+function setSortInfo(fields, schemaInfo) {
+    if (fields.length > 1) {
+        // too many keys
+        throw new Error('Cannot have more than one key in sort object');
```
andreachild, line 491: The apollo/app sync UI doesn't help to guide the user to create multiple sort entries instead of using multiple keys in one entry. Can we change this error message to say something like 'Cannot have more than one field in a single sort object. Please use multiple single-field sort objects instead.'

templates/JSResolverOCHTTPS.js
```diff
@@ -498,9 +515,10 @@ function createQueryFunctionMatchStatement(obj, matchStatements, querySchemaInfo
         const argsAndWhereClauses = extractQueryArgsAndWhereClauses(selection.arguments, querySchemaInfo);
         const queryArgs = argsAndWhereClauses?.queryArguments.length > 0 ? `{${argsAndWhereClauses.queryArguments.join(',')}}` : '';
         const whereClause = argsAndWhereClauses?.whereClauses.length > 0 ? ` WHERE ${argsAndWhereClauses.whereClauses.join(' AND ')}` : '';
+        const sortClause = querySchemaInfo.argOrderBy.length > 0 ? ` ORDER BY ${querySchemaInfo.argOrderBy.map(orderBy => `${querySchemaInfo.pathName}.${orderBy.field} ${orderBy.direction}`).join(', ')}` : '';
```
andreachild, line 518: I did some manual testing and sorting doesn't work for the ID field, which is a special scenario because the id is not a node property but is instead a value from the ID function. 

Currently the generated query looks something like this:

```
ORDER BY getNodeAirports_Airport.country ASC, getNodeAirports_Airport._id DESC
```

but I think it needs to be like this:

```
ORDER BY getNodeAirports_Airport.country ASC, ID(getNodeAirports_Airport) DESC
```

src/test/templates/JSResolverOCHTTPS.test.js
```diff
@@ -200,6 +200,51 @@ test('should resolve app sync event with nested edge filters and variables', ()
...
+            ' WITH getNodeAirports_Airport' +
+            ' ORDER BY getNodeAirports_Airport.desc ASC, getNodeAirports_Airport.code DESC, getNodeAirports_Airport.city DESC\n' +
+            'RETURN collect({desc: getNodeAirports_Airport.`desc`, code: getNodeAirports_Airport.`code`, city: getNodeAirports_Airport.`city`})',
+        parameters: {},
+        language: 'opencypher',
+        refactorOutput: null
+    });
+});
+
+test('should resolve app sync event with nested sort arguments', () => {
```
andreachild, line 221: Can you please add a test that uses sort with arguments from variables?

Example query:
```query GetNodeAirports($options: Options, $sort: [AirportSort!], $nestedOptions: Options, $nestedSort: [AirportSort!]) {
  getNodeAirports(options: $options, sort: $sort) {
    _id
    city
    code
    country
    airportRoutesIn(options: $nestedOptions, sort: $nestedSort) {
      _id
      city
      code
      country
    }
  }
}
```
With variables:
```
{
  "options": {
    "limit": 1
  },
  "sort": [
    {
      "country": "ASC"
    },
    {
      "city": "ASC"
    }
  ],
  "nestedOptions": {
    "limit": 1
  },
  "nestedSort": [
    {
      "country": "DESC"
    },
    {
      "code": "DESC"
    }
  ]
}
```

You can use the test 'should resolve app sync event with nested edge filters and variables' as an example of how to set it up.

src/test/templates/JSResolverOCHTTPS.test.js
```diff
@@ -200,6 +200,86 @@ test('should resolve app sync event with nested edge filters and variables', ()
...
+        refactorOutput: null
+    });
+});
+
+test('should resolve app sync event with nested sort arguments and variables', () => {
+    const result = resolveGraphDBQueryFromAppSyncEvent({
+        field: 'getNodeAirports',
+        arguments: { options: { limit: 1 }, sort: [ { country: 'ASC' }, { city: 'ASC' } ] },
+        variables: {
+            options: { limit: 1 },
```
andreachild, line 253: It doesn't look like the `options` or `sort` variables are used in the query?

andreachild (approved):

PR #108, opened 2025-06-13
Fix resolution of nested variables in selection set arguments

andreachild:
Description of changes:
Missing resolution of nested variables in query selection set arguments is now fixed by recursively iterating through different types of value nodes.

### DIFF
--- a/templates/JSResolverOCHTTPS.js
+++ b/templates/JSResolverOCHTTPS.js
@@ -979,9 +979,7 @@ function replaceVariableInValue(valueNode, variables = {}) {
     if (valueNode.kind === 'ObjectValue' && valueNode.fields) {
         // replace any variables in the object value fields
         valueNode.fields.forEach(field => {
-            if (field.value) {
-                replaceVariableInValue(field.value, variables);
-            }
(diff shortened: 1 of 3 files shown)


templates/JSResolverOCHTTPS.js
```diff
@@ -917,18 +917,59 @@ function convertToValueNode(value) {
...
+        // replace the variable with actual value
+        Object.assign(valueNode, convertToValueNode(variables[valueNode.name.value]));
+        return;
+    }
+    
+    if (valueNode.kind === 'ObjectValue' && valueNode.fields) {
+        // replace any variables in the object value fields
+        valueNode.fields.forEach(field => {
+            if (field.value) {
+                replaceVariableInValue(field.value, variables);
```
Cole-Greer, line 958: Please correct me if I'm wrong here but is this code reliant on `field.value` being passed-by-reference as the recursive call appears to be reassigning it in-place?

If so is there any possibility that `field.value` might be something like a number or string, and thus passed-by-value and not properly updated?

andreachild: I've removed the unnecessary `field.value` check

PR #118, opened 2025-07-07
Fix: Invalid schema generation when AWS AppSync scalars are used in input schema

flora-jin:
Invalid schema was generated if AWS AppSync scalars (AWSDate, AWSTime, AWSEmail, etc...) were referenced in an input schema. Updated code to recognize AWS AppSync scalars as scalar types to ensure that generated schema does not correct them as edges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ You might want to use it as is, or change it. Below are ways to change it.
 ### Please no mutations in my schema
 
 In case you don't want your Graph API having the option of updating your
-database data, add the the CLI option `--output-schema-no-mutations`.
+database data, add the CLI option `--output-schema-no-mutations`.
 
 ### @alias
 
@@ -510,6 +510,34 @@ resolvers, the absence of the custom scalar resolvers will bypass important data
 validation for fields using custom scalar types.
 See [Apollo Custom Scalars](https://www.apollographql.com/docs/apollo-server/schema/custom-scalars)
 for more details on how to attach custom scalar resolvers with Apollo Server.
+
+### AWS AppSync Scalars
+
+AWS AppSync provides a set of extended scalar types that go beyond the default GraphQL 
+scalars like `String`, `Int`, and `Boolean`. 
+
+Refer to 
+[AWS AppSync scalars](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html#graph-ql-aws-appsync-scalars)
+for more details on each of the scalar types.
+
+These scalar types can be used in your schema (specified via `--input-schema-file`) just
+like any other standard scalar when working with **AWS AppSync**. You simply reference 
+them in your schema, and AppSync handles the parsing and validation automatically. 
+For example, you can define a field as an `AWSEmail` or `AWSDateTime` without needing to 
+implement any custom logic as AppSync will enforce the correct format at runtime.
+
+```graphql
+type User {
+  id: ID!
+  email: AWSEmail!
+  createdAt: AWSDateTime!
+  metadata: AWSJSON
+}
+```
+
+However, it's important to note that these scalars are specific to AppSync and are not 
+part of the official GraphQL specification. Thus, these scalars types are **not automatically
+recognized by the Apollo Server** as it is with AppSync.
 
 # Known limitations
 
--- a/doc/todoExample.md
+++ b/doc/todoExample.md
@@ -2,7 +2,7 @@
 
 You can start from a GraphQL schema without directives and an empty Neptune
 database. The utility will inference directives, input, queries and mutations,
-and create the the GraphQL API. Then, you can use GraphQL to create, mutate and
+and create the GraphQL API. Then, you can use GraphQL to create, mutate and
 query the data stored in a Neptune database without the need to know how to use
 a graph query language.
 
@@ -23,6 +23,7 @@ type Todo {
 type Comment {
     content: String
 }
+
 ```
 
 Let's now run this schema through the utility and create the GraphQL API in AWS
@@ -39,7 +40,7 @@ neptune-for-graphql \
 
 The utility created a new file in the *output* folder called
 *TodoExample.source.graphql*, and the GraphQL API in AppSync. As you can see
-below, the utility inferenced:
+below, the utility inferences:
 
 - In the type *Todo* it added *@relationship* for a new type *CommentEdge*. This
   is instructing the resolver to connect *Todo* to *Comment* using a graph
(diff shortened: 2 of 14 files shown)


src/test/schemaModelValidator.test.js
```diff
@@ -101,6 +101,28 @@ describe('validatedSchemaModel', () => {
         const userRoleField = userInput.fields.find(field => field.name.value === 'role');
         expect(userRoleField.type.name.value).toEqual('Role');
     });
+
+    test('should allow AWS AppSync scalars as field types', () => {
```
xShinnRyuu, line 105: Should there also be tests for validating the other AWS AppSync scalar types that exist?

README.md
```diff
@@ -216,7 +216,16 @@ You might want to use it as is, or change it. Below are ways to change it.
 ### Please no mutations in my schema
 
 In case you don't want your Graph API having the option of updating your
-database data, add the the CLI option `--output-schema-no-mutations`.
+database data, add the CLI option `--output-schema-no-mutations`.
+
+### Using AWS AppSync Scalars
+If you wish to use the [AWS AppSync scalars](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html#graph-ql-aws-appsync-scalars),
```
andreachild, line 222: I had tried this a couple weeks ago and from what I recall declaring AWS scalars manually in the schema would result in a warning or error displayed in App Sync.

samples/todo.schema.graphql
```diff
@@ -6,6 +7,7 @@ type Todo {
     status: String
     comments: [Comment]
     bestComment: Comment
+    email: AWSEmail
```
andreachild, line 10: Same comment as https://github.com/aws/amazon-neptune-for-graphql/pull/118/files#r2193510692

src/test/user-group.graphql
```diff
@@ -1,9 +1,11 @@
+scalar AWSEmail
+
 type User {
     userId: ID! @id
     firstName: String
     lastName: String
     role: Role
-    email: EmailAddress
```
andreachild, line 6: This should not be replaced with `AWSEmail` as the test is checking that non-aws custom scalars can still be included in a graphQL schema. I would suggest adding a new field with a different type of AWS custom scalar instead.

README.md
flora-jin: Oh maybe I wasn't clear in the documentation but I meant that the input schema for the utility would require the scalar declaration at the top. So if users add an AWS AppSync scalar to an input schema that goes in the utility, it would require the declaration. Not in AppSync where it has been already deployed.

flora-jin: It is but when I declare it in the input schema and then deploy to AppSync it doesn't generate the error that is seen above.

src/test/schemaModelValidator.test.js
flora-jin: I guess I could but I was just wondering if it would have repetitive as the process is the same for all of the AWS AppSync scalars so I assumed one example would cover all. But now that I think of it I guess it wouldn't hurt to add a separate AWS type that only has AWS fields 🧐

doc/todoExample.md
```diff
@@ -12,18 +12,32 @@ Includes two types: *Todo* and *Comment*. The *Todo* has a field *comments* as
...
 ```graphql
+scalar AWSEmail
+
 type Todo {
     name: String
     description: String
     priority: Int
     status: String
     comments: [Comment]
+    email: AWSEmail
```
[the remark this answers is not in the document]
flora-jin: Would AWSDate be better as a dueDate for the Todo node ?

src/test/airports.customized.graphql
```diff
@@ -54,27 +63,63 @@ _id: ID @id
...
+  emergencyLine: AWSPhone
+  gatewayIp: AWSIPAddress
 }
 
 input CountrySort {
   _id: SortingDirection
   type: SortingDirection
   code: SortingDirection
   desc: SortingDirection
+  foundingDate: SortingDirection
```
andreachild, line 114: Have you verified with app sync that using fields that have customized aws scalar value types in sort works as expected?

flora-jin: Yes the sort works great for me !

doc/todoExample.md
```diff
@@ -18,11 +18,13 @@ type Todo {
     priority: Int
     status: String
     comments: [Comment]
+    dueDate: AWSDate
```
andreachild, line 21: Since the aws scalars only work with AWS app sync and not Apollo I would hesitate to add it to this example. Instead, consider adding a new section for AWS custom scalars under the existing 'Using Custom Scalars with Apollo Server'.

test/testLib.js
```diff
@@ -143,6 +143,7 @@ async function testResolverQueriesResults(resolverFile, queriesReferenceFolder,
     const resolverModule = await loadResolver(resolverFile);
     const schemaDataModelJSON = await decompressGzipToString(schemaFile);
     let schemaModel = JSON.parse(schemaDataModelJSON);
+    injectAwsScalarDefinitions(schemaModel);
```
andreachild, line 146: Curious why this change was needed? I didn't think any of the integration tests used aws scalars?

flora-jin: I think I added it because I just searched for all cases that parsed the resolver as technically the aws scalars should always be supported for appsync so wanted to avoid future confusion just in case. But I guess for now it is unnecessary for integration tests.

doc/todoExample.md
andreachild: Apollo custom scalars are already covered under documentation header 'Using Custom Scalars with Apollo Server'.

README.md
```diff
@@ -511,6 +511,34 @@ validation for fields using custom scalar types.
...
+
+AWS AppSync provides a set of extended scalar types that go beyond the default GraphQL 
+scalars like `String`, `Int`, and `Boolean`. 
+
+Refer to 
+[AWS AppSync scalars](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html#graph-ql-aws-appsync-scalars)
+for more details on each of the scalar types.
+
+These scalar types can be used in you schema (specified via `--input-schema-file`) just
+like any other standard scalar when working with **AWS AppSync**. You simply declare
```
andreachild, line 524:
```suggestion
like any other standard scalar when working with **AWS AppSync**. You simply reference 
```

doc/todoExample.md
```diff
@@ -114,6 +115,7 @@ input TodoUpdateInput {
     description: String
     priority: Int
     status: String
+    dueDate: AWSDate
```
andreachild, line 118: Remove from example since it's aws specific?

flora-jin: Whoops sorry forgot to remove them from the inputs ! Thank you for the check

andreachild (approved):

PR #120, opened 2025-07-18
feat: Added prefix option for query and mutation names

flora-jin:
Added an option to add a custom prefix to generated query and mutation names

- added new --query-prefix option
- value is added to the beginning of generated query names
ex) --query-prefix airport_ will generate airport_getContinent
- added new --query-prefix option to list outputted when user enters -h

- added new --mutation-prefix option
- value is added to the beginning of generated mutation names
ex) --mutation-prefix set to test_ will generate test_createNodeAirport
- added new --mutation-prefix option to list outputted when user enters -h

Added unit tests for changed files to check that the generated schema outputs the correct prefix on query and mutation names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ permissions and limitations under the License.
   `limit` ([#102](https://github.com/aws/amazon-neptune-for-graphql/pull/102))
 * Added support for queries with
   sorting ([#105](https://github.com/aws/amazon-neptune-for-graphql/pull/105))
+* Added options to prefix query and mutation names ([#120](https://github.com/aws/amazon-neptune-for-graphql/pull/120))
 
 ### Improvements
 
--- a/doc/cliReference.md
+++ b/doc/cliReference.md
@@ -94,6 +94,14 @@ database you can edit a graphdb schema as use it as input.
 The Neptune database endpoint from which the utility extract the graphdb schema.
 Format: `host:port`
 
+`--query-prefix`
+<br>
+Optional prefix to add to generated query names in the GraphQL schema.
+
+`--mutation-prefix`
+<br>
+Optional prefix to add to generated mutation names in the GraphQL schema.
+
 ## Output options
 
 `--output-folder-path <value>, -o <value>`
--- a/src/graphdb.js
+++ b/src/graphdb.js
@@ -125,7 +125,21 @@ function formatProperties(properties = [], typeOverrides = new Map()) {
     }).join('');
 }
 
-function graphDBInferenceSchema (graphdbSchema, addMutations) {
+/**
+ * Generates a GraphQL schema for a graph database based on provided node and edge structures.
+ *
+ * This function takes a graph database schema definition and transforms it into a complete GraphQL schema
+ * with types, inputs, queries, and optionally mutations. It handles node and edge relationships,
+ * property definitions, filtering, sorting, and pagination.
+ *
+ * @param {string} graphdbSchema - A JSON string representing the graph database schema with nodeStructures and edgeStructures
+ * @param {boolean} addMutations - Whether to include mutation operations in the generated schema
+ * @param {string} queryPrefix='' - Prefix to add to all query operation names
+ * @param {string} mutationPrefix='' - Prefix to add to all mutation operation names
+ *
+ * @returns {string} A complete GraphQL schema definition as a string
+ */
+function graphDBInferenceSchema (graphdbSchema, { addMutations = true, queryPrefix = '', mutationPrefix = ''} = {}) {
     let r = '';
     const gdbs = JSON.parse(graphdbSchema);
 
@@ -319,8 +333,8 @@ function graphDBInferenceSchema (graphdbSchema, addMutations) {
     r += `type Query {\n`;
     gdbs.nodeStructures.forEach(node => {
         let nodeCase = toPascalCase(cleanseLabel(node.label));
-        r += `\tgetNode${nodeCase}(filter: ${nodeCase}Input): ${nodeCase}\n`;
-        r += `\tgetNode${nodeCase}s(filter: ${nodeCase}Input, options: Options, sort: [${nodeCase}Sort!]): [${nodeCase}]\n`;
+        r += `\t${queryPrefix}getNode${nodeCase}(filter: ${nodeCase}Input): ${nodeCase}\n`;
+        r += `\t${queryPrefix}getNode${nodeCase}s(filter: ${nodeCase}Input, options: Options, sort: [${nodeCase}Sort!]): [${nodeCase}]\n`;
     });
     r += '}\n\n';
 
@@ -329,9 +343,9 @@ function graphDBInferenceSchema (graphdbSchema, addMutations) {
         r += `type Mutation {\n`;
         gdbs.nodeStructures.forEach(node => {
             let nodeCase = toPascalCase(cleanseLabel(node.label));
-            r += `\tcreateNode${nodeCase}(input: ${nodeCase}CreateInput!): ${nodeCase}\n`;
-            r += `\tupdateNode${nodeCase}(input: ${nodeCase}UpdateInput!): ${nodeCase}\n`;
-            r += `\tdeleteNode${nodeCase}(_id: ID!): Boolean\n`;
+            r += `\t${mutationPrefix}createNode${nodeCase}(input: ${nodeCase}CreateInput!): ${nodeCase}\n`;
+            r += `\t${mutationPrefix}updateNode${nodeCase}(input: ${nodeCase}UpdateInput!): ${nodeCase}\n`;
+            r += `\t${mutationPrefix}deleteNode${nodeCase}(_id: ID!): Boolean\n`;
         });    
 
         gdbs.edgeStructures.forEach(edge => {
@@ -341,12 +355,12 @@ function graphDBInferenceSchema (graphdbSchema, addMutations) {
                 let edgeCase = toPascalCase(cleanseLabel(edge.label));
 
                 if (edge.properties.length > 0) {               
-                    r += `\tconnectNode${fromCase}ToNode${toCase}Edge${edgeCase}(from_id: ID!, to_id: ID!, edge: ${edgeCase}Input!): ${edgeCase}\n`;
-                    r += `\tupdateEdge${edgeCase}From${fromCase}To${toCase}(from_id: ID!, to_id: ID!, edge: ${edgeCase}Input!): ${edgeCase}\n`;
+                    r += `\t${mutationPrefix}connectNode${fromCase}ToNode${toCase}Edge${edgeCase}(from_id: ID!, to_id: ID!, edge: ${edgeCase}Input!): ${edgeCase}\n`;
+                    r += `\t${mutationPrefix}updateEdge${edgeCase}From${fromCase}To${toCase}(from_id: ID!, to_id: ID!, edge: ${edgeCase}Input!): ${edgeCase}\n`;
                 } else {
-                    r += `\tconnectNode${fromCase}ToNode${toCase}Edge${edgeCase}(from_id: ID!, to_id: ID!): ${edgeCase}\n`;
+                    r += `\t${mutationPrefix}connectNode${fromCase}ToNode${toCase}Edge${edgeCase}(from_id: ID!, to_id: ID!): ${edgeCase}\n`;
                 }
-                r += `\tdeleteEdge${edgeCase}From${fromCase}To${toCase}(from_id: ID!, to_id: ID!): Boolean\n`;
+                r += `\t${mutationPrefix}deleteEdge${edgeCase}From${fromCase}To${toCase}(from_id: ID!, to_id: ID!): Boolean\n`;
             });
         });
         r += '}\n\n';
--- a/src/schemaModelValidator.js
+++ b/src/schemaModelValidator.js
@@ -15,7 +15,6 @@ import { GraphQLID, print } from 'graphql';
 import {gql} from 'graphql-tag'
 import { loggerInfo, yellow } from "./logger.js";
 
-let quiet = false;
 // TODO change variables to local scope instead of global so this module can be used against multiple schemas
 const typesToAdd = [];
 const queriesToAdd = [];
@@ -137,7 +136,7 @@ function injectChanges(schemaModel) {
 }
 
 
-function addNode(def) {
+function addNode(def, { queryPrefix = '', mutationPrefix = ''} = {}) {
     let name = def.name.value;
     const idField = getIdFieldWithDirective(def);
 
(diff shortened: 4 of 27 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ permissions and limitations under the License.
   `limit` ([#102](https://github.com/aws/amazon-neptune-for-graphql/pull/102))
 * Added support for queries with
   sorting ([#105](https://github.com/aws/amazon-neptune-for-graphql/pull/105))
+* Added option to prefix query and mutation names ([#120](https://github.com/aws/amazon-neptune-for-graphql/pull/120))
```
kmcginnes, line 31:
```suggestion
* Added options to prefix query and mutation names ([#120](https://github.com/aws/amazon-neptune-for-graphql/pull/120))
```

src/graphdb.js
```diff
@@ -125,9 +125,13 @@ function formatProperties(properties = [], typeOverrides = new Map()) {
     }).join('');
 }
 
-function graphDBInferenceSchema (graphdbSchema, addMutations) {
+function graphDBInferenceSchema (graphdbSchema, addMutations, options = {}) {
     let r = '';
     const gdbs = JSON.parse(graphdbSchema);
+    const {
+        inputQueryPrefix = '',
+        inputMutationPrefix = ''
```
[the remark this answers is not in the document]
flora-jin: Will change for schemaModelValidator file as well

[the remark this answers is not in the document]
flora-jin: If preferred I can change the variable name in main.js where the variable is first initialized to queryPrefix but most of the variables accepted by the user starts with input in main.js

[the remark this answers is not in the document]
kmcginnes: You could rename it at the call sites:

```js
        inputGraphQLSchema = graphDBInferenceSchema(inputGraphDBSchema, outputSchemaMutations, {
            queryPrefix: inputQueryPrefix,
            mutationPrefix: inputMutationPrefix
        });
```

Then the variable in `main.js` continues to be `inputQueryPrefix` and any where accessing the `options` would be `queryPrefix.

But I wouldn't waste your time. I'd much rather see a revamp of the options parsing logic. Rather than a long list of variables that are set in the module scope, I'd love to see a function that returns an `options` object with all options parsed or their default values. Then that object is passed to all places that need it. You could even provide some higher order functions so that not every function needs it passed. But that's not for this PR.

kmcginnes (approved):

src/test/schemaModelValidator.test.js
```diff
@@ -10,7 +10,11 @@ describe('validatedSchemaModel', () => {
         loggerInit('./output', false, 'silent');
 
         const schema = readFileSync('./src/test/user-group.graphql');
-        model = validatedSchemaModel(schemaParser(schema));
+        model = validatedSchemaModel(schemaParser(schema), {
+            inputQueryPrefix: 'userGroupQuery_',
```
andreachild, line 14: Please add separate test(s) for these prefixes because otherwise there is no longer test coverage for the more common scenario of not having any prefixes.

Ideally there should also be test(s) for if only one prefix is provided and not both.

src/test/graphdb.test.js
```diff
@@ -25,8 +25,11 @@ test('should output airport schema', async () => {
     expect(actual).toBe(expected);
 });
 
-test('should correctly generate mutation input types after outputting airport schema', async () => {
-    const actual = await inferGraphQLSchema('./src/test/airports-neptune-schema.json', { addMutations: true });
+test('should correctly generate mutation input types with query and mutation prefixes after outputting airport schema', async () => {
+    const actual = await inferGraphQLSchema('./src/test/airports-neptune-schema.json', { addMutations: true }, {
+        inputQueryPrefix: 'airportsQuery_',
```
andreachild, line 30: Please add separate test(s) for these prefixes because otherwise there is no longer test coverage for the more common scenario of not having any prefixes.

Ideally there should also be test(s) for if only one prefix is provided and not both.

andreachild (commented):
Please also make appropriate changes to `cliReference.md` since there are new options.

src/graphdb.js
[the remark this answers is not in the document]
kmcginnes: > @kmcginnes these are brand new options so no renaming would be necessary

The only reason I commented here was because if @flora-jin accepted your suggested change the app would break. Further changes would be necessary to change those names.

Maybe [this TSPlayground](https://www.typescriptlang.org/play/?#code/PTAEFMA8EMFsAcA25QEsB28CuAXAzgFADGA9unjmprgIpbgBOAngAoPgBmqkoAvKACIOJEgIDcxMhSrYcAWVzQcqMm07c+ggEbQG4ggRCgiWBu3SUS8ZVNB4AFtHjhJ5ShlmaA3gVAza9MxqXJAANL7+8oo26MHc4QC+EoZg5nDgACagVjF4do7OrtIeuABK4OjpWfw+fgCOgazsIQBckXSMTephEbDRKrHN3G0lUThKA3E9SQZGpORYsBgA5qAcWOhEMaAA7vaMKDkDeboo8NB4eJlUBOub2zoAXgAUR1Jt0OhMAJSgtcZSSj-PyjDpBIY8fgCAThPwg6hjCaqCGaaGwuENTpTVEwiJ+PrjGLYqG4vwJTRvcgSPzzPAkZAAOkQJGWzwEABFwBQGFgtqZrgA3aCIeh4Fowv549qNKbo+GyBSEyYQuWgTHg7qqglIwaaiIJb4SBKzMAAUUg4BMOBQOH2oBwOxIdiIFV0KkIRXp4CZLLZpQ26BWu1QtogkFQFCDlLyDic3oEhoIT2eo0TnsZzNZAn96ED6FWOxD9lAaVg12j+TjDITEmTo3KlTLGUTQA) will make what I'm trying to say more clear. Check the output when you run it.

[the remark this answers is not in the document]
kmcginnes: > @kmcginnes these are brand new options so no renaming would be necessary

If the currently suggested code changes are committed without additional changes, the app will break. This particular function would ignore the option values passed to it and the values will always be an empty string.

Perhaps this [TS Playground](https://www.typescriptlang.org/play/?#code/PTAEFMA8EMFsAcA25QEsB28CuAXAzgFADGA9unjmprgIpbgBOAngAoPgBmqkoAvKACIOJEgIDcxMhSrYcAWVzQcqMm07c+ggEbQG4ggRCgiWBu3SUS8ZVNB4AFtHjhJ5ShlmaA3gVAza9MxqXJAANL7+8oo26MHc4QC+EoZg5nDgACagVjF4do7OrtIeuABK4OjpWfw+fgCOgazsIQBckXSMTephEbDRKrHN3G0lUThKA3E9SQZGpORYsBgA5qAcWOhEMaAA7vaMKDkDeboo8NB4eJlUBOub2zoAXgAUR1Jt0OhMAJSgtcZSSj-PyjDpBIY8fgCAThPwg6hjCaqCGaaGwuENTpTVEwiJ+PrjGLYqG4vwJTRvcgSPzzPAkZAAOkQJGWzwEABFwBQGFgtqZrgA3aCIeh4Fowv549qNKbo+GyBSEyYQuWgTHg7qqglIwaaiIJb4SBKzMAAUUg4BMOBQOH2oBwOxIdiIFV0KkIRXp4CZLLZpQ26BWu1QtogkFQFCDlLyDic3oEhoIT2eo0TnsZzNZAn96ED6FWOxD9lAaVg12j+TjDITEmTo3KlTLGUTQA) will help explain what I mean a bit better. Pay attention to the output of both scenarios.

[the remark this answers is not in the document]
andreachild: Ah yes I know what you mean now. The suggestion would not work as-is. It would have to be reworked without the destructuring.

doc/cliReference.md
```diff
@@ -94,6 +94,14 @@ database you can edit a graphdb schema as use it as input.
 The Neptune database endpoint from which the utility extract the graphdb schema.
 Format: `host:port`
 
+`--query-prefix`
+<br>
+Adds a prefix to query names
```
andreachild, line 99:
```suggestion
Optional prefix to add to generated query names in the GraphQL schema.
```

doc/cliReference.md
```diff
@@ -94,6 +94,14 @@ database you can edit a graphdb schema as use it as input.
...
+
+`--mutation-prefix`
+<br>
+Adds a prefix to mutation names
```
andreachild, line 103:
```suggestion
Optional prefix to add to generated mutation names in the GraphQL schema.
```

src/schemaModelValidator.js
```diff
@@ -15,14 +15,22 @@ import { GraphQLID, print } from 'graphql';
...
 
-let quiet = false;
 // TODO change variables to local scope instead of global so this module can be used against multiple schemas
 const typesToAdd = [];
 const queriesToAdd = [];
 const mutationsToAdd = [];
 const enumTypes = [];
 const customScalarTypes = ['AWSDate', 'AWSTime', 'AWSDateTime', 'AWSTimestamp', 'AWSEmail', 'AWSJSON', 'AWSPhone', 'AWSURL', 'AWSIPAddress'];
 
+function clearVars() {
```
andreachild, line 25: Was this function only added for unit testing purposes?

src/schemaModelValidator.js
```diff
@@ -137,9 +145,13 @@ function injectChanges(schemaModel) {
...
 
 
-function addNode(def) {
+function addNode(def, options = {}) {
     let name = def.name.value;
     const idField = getIdFieldWithDirective(def);
+    const {
+        inputQueryPrefix = '',
+        inputMutationPrefix = ''
+    } = options;
```
andreachild, line 154:
```suggestion
function addNode(def, options = {
    queryPrefix: '',
    mutationPrefix: ''
}) {
    let name = def.name.value;
    const idField = getIdFieldWithDirective(def);
```

src/schemaModelValidator.js
```diff
@@ -191,34 +203,38 @@ function addNode(def) {
...
 
     loggerInfo(`Added input type: ${yellow(name+'Input')}`);
     loggerInfo(`Added input type: ${yellow(name+'CreateInput')}`);
     loggerInfo(`Added input type: ${yellow(name+'UpdateInput')}`);
-    loggerInfo(`Added query: ${yellow('getNode' + name)}`);
-    loggerInfo(`Added query: ${yellow('getNode' + name + 's')}`);
-    loggerInfo(`Added mutation: ${yellow('createNode' + name)}`);
-    loggerInfo(`Added mutation: ${yellow('updateNode' + name)}`);
-    loggerInfo(`Added mutation: ${yellow('deleteNode' + name)}`);
+    loggerInfo(`Added query: ${yellow(inputQueryPrefix + 'getNode' + name)}`);
```
andreachild, line 217: These names will change again soon so might as well refactor them now so you don't need to change the logging lines again:

```
// Create query
const getSingleQueryName = `${inputQueryPrefix}getNode${name}`;
queriesToAdd.push(`${getSingleQueryName}(filter: ${name}Input): ${name}\n`);
const getMultipleQueryName = `${inputQueryPrefix}getNode${name}s`;
queriesToAdd.push(`${getMultipleQueryName}(filter: ${name}Input, options: Options, sort: [${name}Sort!]): [${name}]\n`);

// Create mutation
const createMutationName = `${inputMutationPrefix}createNode${name}`;
mutationsToAdd.push(`${createMutationName}(input: ${name}CreateInput!): ${name}\n`);
const updateMutationName = `${inputMutationPrefix}updateNode${name}`;
mutationsToAdd.push(`${updateMutationName}(input: ${name}UpdateInput!): ${name}\n`);
const deleteMutationName = `${inputMutationPrefix}deleteNode${name}`;
mutationsToAdd.push(`${deleteMutationName}(${print(idFieldToInputValue(idField))}): Boolean\n`);

loggerInfo(`Added input type: ${yellow(name+'Input')}`);
loggerInfo(`Added input type: ${yellow(name+'CreateInput')}`);
loggerInfo(`Added input type: ${yellow(name+'UpdateInput')}`);
loggerInfo(`Added query: ${yellow(getSingleQueryName)}`);
loggerInfo(`Added query: ${yellow(getMultipleQueryName)}`);
loggerInfo(`Added mutation: ${yellow(createMutationName)}`);
loggerInfo(`Added mutation: ${yellow(updateMutationName)}`);
loggerInfo(`Added mutation: ${yellow(deleteMutationName)}`);
```

src/test/airports.customized.graphql
```diff
@@ -291,17 +291,23 @@ getNodeCountrys(filter: CountryInput, options: Options, sort: [CountrySort!]): [
...
 getNodeAirport(filter: AirportInput): Airport
+airportQuery_getNodeAirport(filter: AirportInput): Airport
 getNodeAirports(filter: AirportInput, options: Options, sort: [AirportSort!]): [Airport]
 getAirportWithGremlin(code: String): Airport @graphQuery(statement: "g.V().has('airport', 'code', '$code').elementMap()")
 getCountriesCount: Int @graphQuery(statement: "g.V().hasLabel('country').count()")
 }
 
 type Mutation {
 createNodeAirport(input: AirportCreateInput!): Airport
+airportMutation_createNodeAirport(input: AirportCreateInput!): Airport
```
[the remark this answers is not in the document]
flora-jin: This is because I added unit tests for the resolver to confirm that the mutation queries were correctly translated by the resolver with prefixes. So I added some mutations that have prefixes for the resolver to use.

src/schemaModelValidator.js
flora-jin: Yes, the problem I ran into while running the validatedSchemaModel function multiple times was that the global variables were not being emptied so I felt that the refactoring for those variables to be local would be for a separate ticket so temporarily added a function to empty the global variables every time the function validatedSchemaModel is ran.

src/graphdb.js
```diff
@@ -125,9 +125,28 @@ function formatProperties(properties = [], typeOverrides = new Map()) {
...
+ *
+ * @returns {string} A complete GraphQL schema definition as a string
+ */
+function graphDBInferenceSchema (graphdbSchema, addMutations, options = {}) {
     let r = '';
     const gdbs = JSON.parse(graphdbSchema);
+    const {
+        inputQueryPrefix = '',
+        inputMutationPrefix = ''
+    } = options;
```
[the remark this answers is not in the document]
flora-jin: Also, I think the suggested code isn't deconstructing options so the inputQueryPrefix and inputMutationPrefix from main.js is not being stored in queryPrefix or mutationPrefix. For the options object, it would require consistent keys to pass the variables over to the function from the function call.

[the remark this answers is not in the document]
flora-jin: I can remove the deconstructing and just use the options object as is because main.js always has an initialization of the parameters, but I don't know if that would be a safe option as that means the function is unable to be called with a partial options object. For example, if we move addMutations to the options object, in a case like graphdb.test.js file, the unit tests call the graphDBInferenceSchema function with a partial options object where only the inputQueryPrefix variable is initialized. In cases like that, this implementation would break as the included partial options object would override the pre-existing default values and make some parts of the options object undefined. The default values for the options object only works if the user does not include an options object at all. It would break for partial options objects.

src/schemaModelValidator.js
andreachild: In general it's not advisable to add code to non-test files just for testing purposes. It would be nice to refactor schemaModelValidator to not use module-scoped variables but in lieu of doing that you can instead import the schemaModelValidator as needed with fresh module scope using a function in your test file like this (the timestamp will create a fresh entry in the module cache):

```
async function loadSchemaValidator() {
    // Use timestamp to get fresh module
    const timestamp = Date.now();
    return await import(`../schemaModelValidator.js?t=${timestamp}`);
}
```

Then the test can use the function like this:

```
test('should output expected validated schema with query and mutation prefixes', async () => {
    const schemaValidator = await loadSchemaValidator();
    const modelPrefixes = schemaValidator.validatedSchemaModel(schemaParser(schema), {
        inputQueryPrefix: 'userGroupQueryTest_',
        inputMutationPrefix: 'userGroupMutationTest_'
    });
    const actual = schemaStringify(modelPrefixes, true);
    const expected = readFileSync('./src/test/user-group-validated-with-prefixes.graphql', 'utf8');
    expect(actual).toBe(expected);
})

test('should output expected validated schema with mutation prefixes', async () => {
    const schemaValidator = await loadSchemaValidator();
    const modelMutationPrefix = schemaValidator.validatedSchemaModel(schemaParser(schema), {
        inputMutationPrefix: 'userGroupMutation_'
    });
    const actual = schemaStringify(modelMutationPrefix, true);
    const expected = readFileSync('./src/test/user-group-validated-with-mutation-prefix.graphql', 'utf8');
    expect(actual).toBe(expected);
})
```

templates/JSResolverOCHTTPS.js
```diff
@@ -1102,7 +1102,7 @@ function getReturnBlockFromSelection(selection, querySchemaInfo) {
 function resolveGrapgDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
 
     // createNode
-    if (querySchemaInfo.name.startsWith('createNode') && !querySchemaInfo.graphQuery) {
+    if (/^.*createNode/.test(querySchemaInfo.name) && !querySchemaInfo.graphQuery) {
```
andreachild, line 1105: It would be better for performance to just use `querySchemaInfo.name.includes('createNode')` which is essentially the same check as that regex pattern.

```suggestion
    if (querySchemaInfo.name.includes('createNode') && !querySchemaInfo.graphQuery) {
```

templates/JSResolverOCHTTPS.js
```diff
@@ -1164,7 +1164,7 @@ function resolveGrapgDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
     }
 
     // connect
-    if (querySchemaInfo.name.startsWith('connectNode') && querySchemaInfo.graphQuery == null) {
+    if (/^.*connectNode/.test(querySchemaInfo.name) && querySchemaInfo.graphQuery == null) {
```
andreachild, line 1167:
```suggestion
    if (querySchemaInfo.name.includes('connectNode') && querySchemaInfo.graphQuery == null) {
```

templates/JSResolverOCHTTPS.js
```diff
@@ -1182,7 +1182,7 @@ function resolveGrapgDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
     }
 
     // updateEdge
-    if (querySchemaInfo.name.startsWith('updateEdge') && querySchemaInfo.graphQuery == null) {
+    if (/^.*updateEdge/.test(querySchemaInfo.name) && querySchemaInfo.graphQuery == null) {
```
andreachild, line 1185:
```suggestion
    if (querySchemaInfo.name.includes('updateEdge') && querySchemaInfo.graphQuery == null) {
```

templates/JSResolverOCHTTPS.js
```diff
@@ -1209,7 +1209,7 @@ function resolveGrapgDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
     }
 
     // deleteEdge
-    if (querySchemaInfo.name.startsWith('deleteEdge') && querySchemaInfo.graphQuery == null) {
+    if (/^.*deleteEdge/.test(querySchemaInfo.name) && querySchemaInfo.graphQuery == null) {
```
andreachild, line 1212:
```suggestion
    if (querySchemaInfo.name.includes('deleteEdge') && querySchemaInfo.graphQuery == null) {
```

templates/JSResolverOCHTTPS.js
flora-jin: querySchemaInfo.name is createAirport for example. that would be createcreateAirport

flora-jin: But I can change it to includes I think

andreachild: You can look at the other fields in `querySchemaInfo`, the node label should be there because it's used to generate the cypher query which requires the node label. Maybe `returnType`?

test/TestCases/Case01/case02.json
andreachild, line 1: I think it would be better to move this prefix test to the Case05 tests because those are actually deploying an app sync pipeline and you can execute the graphQL query against the app sync API.

See `Case05.02.test.js` to see how you can execute a graphQL query directly against the deployed app sync API.

test/TestCases/Case05/case03.json
andreachild, line 1: Going forward I would like the test case json files with the command line arguments to be descriptive of what is being tested. Can you please change the name of this file to something like `pipeline-with-prefixes.json`?

test/TestCases/Case05/Case05.05.test.js
```diff
@@ -0,0 +1,65 @@
...
+        const unzippedFolder = path.join(outputFolderPath, 'unzipped');
+        const actualFiles = unzipAndGetContents(unzippedFolder, path.join(outputFolderPath, 'AirportsJestPrefixTest.zip'));
+        expect(actualFiles.toSorted()).toEqual(expectedFiles.toSorted());
+
+        // resolver should be using axios http client
+        const fileContent = fs.readFileSync(path.join(unzippedFolder, 'index.mjs'), 'utf8');
+        expect(fileContent).toContain('axios');
+    });
+
+    test('Can query app sync API successfully', async () => {
```
andreachild, line 39:
```suggestion
    test('Can query app sync API successfully with prefixes', async () => {
```

test/TestCases/Case05/Case05.05.test.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(codes).toEqual(['AF', 'AN', 'AS', 'EU', 'NA', 'OC', 'SA']);
+    }, 600000);
+
+    test('Can execute a mutation on app sync API successfully', async () => {
+        const awsResources = JSON.parse(fs.readFileSync(path.join(outputFolderPath, 'AirportsJestPrefixTest-resources.json'), 'utf8'));
+        const apiId = awsResources.AppSyncAPI;
+        const region = awsResources.region;
+
+        const mutation = `mutation {
+            airportsMutationPrefix_createNodeAirport(input: {
```
andreachild, line 54: Can you please also execute a deletion request for this created node so that the graph is left in the same state it started with?

andreachild (approved):

PR #124, opened 2025-07-29
feat: Changed query and mutation names to be less graph-specific

flora-jin:
Updated query and mutation names to be less graph-specific

Queries:
- removed Node from query names ("getNode..." -> "get...")

Mutations:
- removed Node from mutation names ("createNode..." -> "create...")
- updated connect mutation name by removing Node and replacing Edge with Through ("connectNode...ToNode...Edge..." -> "connect...To...Through...")
- updated delete and update mutation name by removing Edge and replacing with Connection ("updateEdge...From...To..." -> "update...ConnectionFrom...To...")

Updated integration and unit tests to follow this change in schema pattern for query and mutations.
Also updated any documentations including previous query and mutation namings.

### DIFF
--- a/samples/airports.source.schema.graphql
+++ b/samples/airports.source.schema.graphql
@@ -234,7 +234,7 @@ input StringScalarFilters {
 }
 
 type Query {
-  getAirportCustom(code: String): Airport
+  getAirportByCode(code: String): Airport
   getAirportConnection(fromCode: String!, toCode: String!): Airport @cypher(statement: "MATCH (:airport{code: '$fromCode'})-[:route]->(this:airport)-[:route]->(:airport{code:'$toCode'})")
   getAirportWithGremlin(code:String): Airport @graphQuery(statement: "g.V().has('airport', 'code', '$code').elementMap()")
   getAirportsWithGremlin: [Airport] @graphQuery(statement: "g.V().hasLabel('airport').elementMap().fold()")    
@@ -251,9 +251,9 @@ type Query {
 }
 
 type Mutation {
-  createAirportCustom(input: AirportInput!): Airport @graphQuery(statement: "CREATE (this:airport {$input}) RETURN this")
+  createAirportWithCypher(input: AirportInput!): Airport @graphQuery(statement: "CREATE (this:airport {$input}) RETURN this")
   addRoute(fromAirportCode:String, toAirportCode:String, dist:Int): Route @graphQuery(statement: "MATCH (from:airport{code:'$fromAirportCode'}), (to:airport{code:'$toAirportCode'}) CREATE (from)-[this:route{dist:$dist}]->(to) RETURN this")
-  deleteAirportCustom(id: ID): Int @graphQuery(statement: "MATCH (this:airport) WHERE ID(this) = '$id' DETACH DELETE this")
+  deleteAirportWithCypher(id: ID): Int @graphQuery(statement: "MATCH (this:airport) WHERE ID(this) = '$id' DETACH DELETE this")
 
   createContinent(input: ContinentCreateInput!): Continent
   updateContinent(input: ContinentUpdateInput!): Continent


kmcginnes:
Technically, this would be a breaking change. Any concerns about that or is the user base small enough not to worry too much?

doc/routesExample.md
```diff
@@ -486,36 +486,36 @@ input StringScalarFilters {
...
-    getNodeCountry(filter: CountryInput): Country
-    getNodeCountrys(filter: CountryInput, options: Options, sort: [CountrySort!]): [Country]
-    getNodeVersion(filter: VersionInput): Version
-    getNodeVersions(filter: VersionInput, options: Options, sort: [VersionSort!]): [Version]
-    getNodeAirport(filter: AirportInput): Airport
-    getNodeAirports(filter: AirportInput, options: Options, sort: [AirportSort!]): [Airport]
+    getContinent(filter: ContinentInput): Continent
+    getContinents(filter: ContinentInput, options: Options, sort: [ContinentSort!]): [Continent]
+    getCountry(filter: CountryInput): Country
+    getCountrys(filter: CountryInput, options: Options, sort: [CountrySort!]): [Country]
```
kmcginnes, line 492: English is so unfriendly to developers 🤦 

Do we want to make any attempt at addressing words that end in "y" when pluralized?

templates/JSResolverOCHTTPS.js
```diff
@@ -1101,8 +1101,8 @@ function getReturnBlockFromSelection(selection, querySchemaInfo) {
 
 function resolveGrapgDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
```
kmcginnes, line 1102:
```suggestion
function resolveGraphDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
```

Misspelled function name

templates/JSResolverOCHTTPS.js
```diff
@@ -1118,75 +1118,81 @@ function resolveGrapgDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
...
-        const nodeName = querySchemaInfo.name + '_' + querySchemaInfo.returnType;
-        const idParam  = `${nodeName}_${idField.name}`;
-        Object.assign(parameters, {[idParam]: nodeID});
-        
-        let returnBlock = `ID(${nodeName})`;
-        if (queryAst.definitions[0].selectionSet.selections[0].selectionSet) {
-            returnBlock = getReturnBlockFromSelection(queryAst.definitions[0].selectionSet.selections[0], querySchemaInfo);
+    // update
+    if (querySchemaInfo.name.includes(`update${querySchemaInfo.returnType}`) && !querySchemaInfo.graphQuery) {
+        const nameWithoutUpdate = querySchemaInfo.name.split('update').pop();
```
kmcginnes, line 1123: Would `nameWithoutUpdate` always be equal to `querySchemaInfo.returnType`?

templates/JSResolverOCHTTPS.js
```diff
@@ -1118,75 +1118,81 @@ function resolveGrapgDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
...
-    // deleteNode
-    if (querySchemaInfo.name.includes('deleteNode') && !querySchemaInfo.graphQuery) {
-        const nodeID = queryAst.definitions[0].selectionSet.selections[0].arguments[0].value.value;
-        const nodeName = querySchemaInfo.name + '_' + querySchemaInfo.returnType;
-        let param  = nodeName + '_' + 'whereId';
-        Object.assign(parameters, {[param]: nodeID});
-        const ocQuery = `MATCH (${nodeName})\nWHERE ID(${nodeName}) = $${param}\nDETACH DELETE ${nodeName}\nRETURN true`;
-        return ocQuery;
+    // delete
+    if (querySchemaInfo.name.includes('delete') && querySchemaInfo.returnType === 'Boolean' && !querySchemaInfo.graphQuery) {
```
kmcginnes, line 1160: Should this conditional check if the name includes `delete${querySchemaInfo.returnType}` similar to the create and update?

src/test/templates/JSResolverOCHTTPS.test.js
```diff
@@ -397,11 +397,11 @@ test('should resolve app sync event with AWS AppSync scalars as field types', ()
 
 // Query0001
 test('should inference query from return type (Query0001)', () => {
-    const result = resolveGraphDBQuery({queryObjOrStr: 'query MyQuery {\n getAirport(code: \"SEA\") {\n city \n }\n}'});
+    const result = resolveGraphDBQuery({queryObjOrStr: 'query MyQuery {\n getAirportByCode(code: \"SEA\") {\n city \n }\n}'});
```
kmcginnes, line 400: I'm struggling to find the reason why this change occurred.

templates/JSResolverOCHTTPS.js
flora-jin: For update queries it should be as the generated schema from the utility generates that way.

src/test/templates/JSResolverOCHTTPS.test.js
```diff
@@ -1045,123 +1045,125 @@ test('should resolve custom mutation with @graphQuery directive and $input param
...
         language: 'opencypher',
         refactorOutput: null
     });
 });
 
-test('should resolve mutation to delete edge between nodes with a prefix', () => {
-    const query = 'mutation DeleteEdgeContainsFromCountryToAirport {\n' +
-        '  airportMutation_deleteEdgeContainsFromCountryToAirport(from_id: \"ee71c547-ea32-4573-88bc-6ecb31942a1e\", to_id: \"99cb3321-9cda-41b6-b760-e88ead3e1ea1\")\n' +
+test('should resolve mutation to delete connection between nodes with a prefix', () => {
+    const query = 'mutation DeleteContainsConnectionFromCountryToAirport {\n' +
```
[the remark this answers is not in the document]
flora-jin: The operation name normally has the first letter capitalized like 
```
mutation MyMutation
```
which is why I believe it follows this convention

andreachild:
> Technically, this would be a breaking change. Any concerns about that or is the user base small enough not to worry too much?

Breaking change is OK as we are implementing a number of breaking changes for the 2.0 release.

doc/routesExample.md
andreachild: I think Q could be used to generate a quick function to pluralize most english words which you can put into a new `pluralize.js` file.

andreachild: > Perhaps we can side-step the complexity of pluralization by changing the first word
Unfortunately `filterCountry` should still probably be `filterCountries` if we were to change it.

templates/JSResolverOCHTTPS.js
andreachild: This check is because there can be mutations for updating a specific node or updating an edge which can both resolve 'true' for the if condition. Examples:
`updateEdgeRouteFromAirportToAirport`
`updateAirport`
`myPrefix_updateAirport`
`myPrefix_updateEdgeRouteFromAirportToAirport`

What if the `if` condition was changed to:
```
if (querySchemaInfo.name.endsWith(`update${querySchemaInfo.returnType}`) && !querySchemaInfo.graphQuery)
```
?

templates/JSResolverOCHTTPS.js
```diff
@@ -1118,75 +1118,81 @@ function resolveGrapgDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
...
+        const nameWithoutDelete = querySchemaInfo.name.split('delete').pop();
+        if (getTypeDefs(nameWithoutDelete).length) {
+            const nodeID = queryAst.definitions[0].selectionSet.selections[0].arguments[0].value.value;
+            const nodeName = querySchemaInfo.name + '_' + querySchemaInfo.returnType;
+            let param  = nodeName + '_' + 'whereId';
+            Object.assign(parameters, {[param]: nodeID});
+            const ocQuery = `MATCH (${nodeName})\nWHERE ID(${nodeName}) = $${param}\nDETACH DELETE ${nodeName}\nRETURN true`;
+            return ocQuery;
+        }
```
andreachild, line 1169: Same comment as https://github.com/aws/amazon-neptune-for-graphql/pull/124/files#r2246557368

test/TestCases/Case01/input/airportsPrefix.source.schema.graphql
andreachild, line 1: I don't see any references or usages of this file?

test/TestCases/Case01/queries/Query0015.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "Mutation: create node",
     "description": "",
-    "graphql": "mutation MyMutation {\n createNodeAirport(input: {code: \"NAX\", city: \"Reggio Emilia\"}) {\n code\n }\n }",
+    "graphql": "mutation MyMutation {\n createAirport(input: {code: \"NAX\", city: \"Reggio Emilia\"}) {\n code\n }\n }",
```
andreachild, line 4: I think this file should be deleted because what I noticed is that this query is causing multiple `Reggio Emilia` airports to be created in the target graph. This test can't perform proper cleanup of test data. We are already validating that create node queries are resolved correctly in the resolver unit test and we have a different integration test which is creating an airport and then deleting it afterwards.

samples/airports.source.schema.graphql
```diff
@@ -234,46 +234,46 @@ input StringScalarFilters {
 }
 
 type Query {
-  getAirport(code: String): Airport
+  getAirportCustom(code: String): Airport
```
andreachild, line 237: Since this is a sample file and not a test file, how about changing the query name to `getAirportByCode` which is more realistic?

andreachild (approved):

PR #125, opened 2025-08-06
fix: Fixed duplicated nodes and edges from multi-label nodes

flora-jin:
There was a bug where multi-label nodes would duplicate edges as the function 
```
findFromAndToLabels(edgeStructure)
```
(which looks at a specific edge type in the graph database to see what kinds of nodes it connects, creating a list of all the 
different from-to label pairs for that relationship) used nested for loops to go through all combinations of fromLabel and toLabel arrays. When nodes had multiple labels, this created duplicate edge directions since each label combination would generate a separate entry in edgeStructure.directions, even if the actual relationship between the node types was the same. 

Also, the function
```
getNodeNames()
```
(which gets all the different node types in the graph database by querying for all nodes and their labels, then adds 
each label to the schema structure) generated duplicated nodes in the generated schema as the original code was processing node labels without checking for duplicates, causing the same node label to be added multiple times 
to schema.nodeStructures.

The fix creates a new empty set every time the function is called to store every processed edge or node in order to check that it has not been previously processed.

### DIFF
--- a/src/NeptuneSchema.js
+++ b/src/NeptuneSchema.js
@@ -148,10 +148,17 @@ async function getNodesNames() {
     loggerInfo('Getting nodes names');
 
     try {
+        const processedLabels = new Set(schema.nodeStructures.map(node => node.label));
+
         response.results.forEach(result => {
-            schema.nodeStructures.push({ label: result['labels(a)'][0], properties: []});
-            loggerDebug('Found Node: ' + yellow(result['labels(a)'][0]), {toConsole: true});
-        });        
+            result['labels(a)'].forEach(label => {
+                if (!processedLabels.has(label)) {
+                    schema.nodeStructures.push({ label: label, properties: []});
+                    processedLabels.add(label);
+                    loggerDebug('Found Node: ' + yellow(label), {toConsole: true});
+                }
+            });
+        });
     }
     catch (e)  {
         loggerError('No nodes found', e);
@@ -181,11 +188,20 @@ async function findFromAndToLabels(edgeStructure) {
     const query = `MATCH (from)-[r:${sanitize(edgeStructure.label)}]->(to) WITH from, to LIMIT $sample RETURN DISTINCT labels(from) as fromLabel, labels(to) as toLabel`;
     loggerDebug(`Retrieving incoming and outgoing labels for edge ${edgeStructure.label} with limit ${SAMPLE}`, {toConsole: true});
     const response = await queryNeptune(query, {sample: SAMPLE});
+    const existingDirections = new Set(
+        edgeStructure.directions.map(dir => `${dir.from}|${dir.to}`)
+    );
+
     for (let result of response.results) {
         for (let fromLabel of result.fromLabel) {
             for (let toLabel of result.toLabel) {
-                edgeStructure.directions.push({from:fromLabel, to:toLabel});
-                loggerDebug('Found edge: ' + yellow(edgeStructure.label) + '  direction: ' + yellow(fromLabel) + ' -> ' + yellow(toLabel), {toConsole: true});
+                const directionKey = `${fromLabel}|${toLabel}`;
+
+                if (!existingDirections.has(directionKey)) {
+                    edgeStructure.directions.push({from: fromLabel, to: toLabel});
+                    existingDirections.add(directionKey);
+                    loggerDebug('Found edge: ' + yellow(edgeStructure.label) + '  direction: ' + yellow(fromLabel) + ' -> ' + yellow(toLabel), {toConsole: true});
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/test/NeptuneSchema.test.js
```diff
@@ -0,0 +1,174 @@
+import { jest } from '@jest/globals';
+import axios from 'axios';
+import { loggerInit } from '../logger.js';
+
+describe('Neptune Schema', () => {
```
andreachild, line 5:
```suggestion
describe('Neptune Schema discovery tests', () => {
```

src/test/NeptuneSchema.test.js
```diff
@@ -0,0 +1,174 @@
...
+                        { fromLabel: ['Developer'], toLabel: ['Manager'] }
+                    ]
+                }
+            },
+            { data: { results: [] } },
+            { data: { results: [] } },
+            { data: { results: [] } },
+            { data: { results: [{ rels: 3 }] } }
+        ];
+        mockAxiosResponses(mockResponses);
```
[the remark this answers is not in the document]
kmcginnes: This should also use a spy instead of a mock

src/test/NeptuneSchema.test.js
```diff
@@ -0,0 +1,174 @@
...
+        const edgeLabels = schema.edgeStructures.map(edge => edge.label).sort();
+        expect(edgeLabels).toEqual(['REPORTS_TO', 'WORKS_FOR']);
+
+        // Restore console.error at the end
+        consoleSpy.mockRestore();
+    });
+});
+
+async function runGetNeptuneSchema() {
+    const {getNeptuneSchema, setGetNeptuneSchemaParameters} = await import(`../NeptuneSchema.js?t=${Date.now()}`);
```
kmcginnes, line 149: You might be able to use `jest.resetModules()` inside your `beforeEach` to get the same effect. That might allow you to import these methods like normal.

I say "might" because I don't ever have module specific state that needs resetting, so I don't typically use this.

src/test/NeptuneSchema.test.js
```diff
@@ -0,0 +1,174 @@
...
+        expect(schema.nodeStructures).toHaveLength(6);
+        const nodeLabels = schema.nodeStructures.map(node => node.label).sort();
+        expect(nodeLabels).toEqual(['Company', 'Developer', 'Employee', 'Manager', 'Person', 'TeamLead']);
+
+        expect(schema.edgeStructures).toHaveLength(2);
+        const edgeLabels = schema.edgeStructures.map(edge => edge.label).sort();
+        expect(edgeLabels).toEqual(['REPORTS_TO', 'WORKS_FOR']);
+
+        // Restore console.error at the end
+        consoleSpy.mockRestore();
```
kmcginnes, line 144: Won't need this if you do the `jest.restoreAllMocks()` in the `beforeEach`.

Although, I'm not sure what happens after the last test runs. You may want to put any of the restoring stuff in an `afterEach` instead.

src/NeptuneSchema.js
```diff
@@ -147,11 +147,18 @@ async function getNodesNames() {
...
+
     try {
         response.results.forEach(result => {
-            schema.nodeStructures.push({ label: result['labels(a)'][0], properties: []});
-            loggerDebug('Found Node: ' + yellow(result['labels(a)'][0]), {toConsole: true});
-        });        
+            result['labels(a)'].forEach(label => {
+                if (!processedNodeLabels.has(label)) {
+                    processedNodeLabels.add(label);
+                    schema.nodeStructures.push({ label: label, properties: []});
```
kmcginnes, line 157: This could lead to duplicate entries in `schema.nodeStructures` if some other function adds to the `nodeStructures` array.

The `processedNodeLabels` is only protecting against duplicates within this `response` data. It doesn't protect `schema.nodeStructures` directly.

kmcginnes: Your approach does solve the concern, but it also is less performant. You might be able to get the best of both worlds by pre-filling the original `Set` based logic with the contents of `schema.nodeStructures` mapped to whatever the key shape is.

But any option should be good.

src/test/NeptuneSchema.test.js
```diff
@@ -0,0 +1,171 @@
+import { jest } from '@jest/globals';
+import axios from 'axios';
+import { loggerInit } from '../logger.js';
+
+describe('Neptune Schema discovery tests', () => {
+    beforeAll(() => {
+        loggerInit('./src/test/output', true, 'fatal');
+    });
+
+    beforeEach(() => {
```
kmcginnes, line 10: You should use `afterEach` for `jest.restoreAllMocks()` like this example from the [Jest docs](https://jestjs.io/docs/jest-object#jestspyonobject-methodname-accesstype)

andreachild (approved):

kmcginnes (approved):

PR #127, opened 2025-08-18
bug: Accommodated for special characters in edge labels to ensure correct Cypher translation

flora-jin:
Queries would fail if an edge label contained a special character such as a colon, period, or space as it is invalid formatting for Cypher. Thus, back ticks were added to ensure that special characters could be included as label names without crashing from incorrect Cypher translations.

Added a unit test for the resolver to ensure backticks are added correctly in cypher translations from queries with special characters in edge labels. Also changed existing resolver unit tests to accommodate for this change.

### DIFF
--- a/src/test/airports.customized.graphql
+++ b/src/test/airports.customized.graphql
@@ -8,15 +8,15 @@ type Continent @alias(property: "continent") {
   type: String
   code: String
   desc: String
-  airportContainssOut(filter: AirportInput, options: Options): [Airport] @relationship(edgeType: "contains", direction: OUT)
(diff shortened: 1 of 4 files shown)


src/test/airports.customized.graphql
```diff
@@ -267,6 +319,10 @@ input RouteInput {
 dist: Int
 }
 
+type PulledFrom @alias(property:"pulled:From") {
```
andreachild, line 322: If this schema had been generated by the utility the type name would be something like `Pulled_cn_From`.

andreachild (approved):

PR #136, opened 2025-09-08
Fixed graphdb.js to process ONE-ONE edge types

andreachild:
Fixed graphdb.js to process ONE-ONE edge types by adding appropriate `@relationship` directives, allowing resulting graphQL schema to traverse along these edges.

### DIFF
--- a/src/graphdb.js
+++ b/src/graphdb.js
@@ -195,7 +195,7 @@ function graphDBInferenceSchema (graphdbSchema, { addMutations = true, queryPref
                     if (direction.relationship === 'ONE-MANY') {
                         r += `\t${toCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${toCase}Input, options: Options, sort: [${toCase}Sort!]): [${toCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
                     }
-                    if (direction.relationship === 'MANY-ONE') {
+                    if (direction.relationship === 'MANY-ONE' || direction.relationship === 'ONE-ONE') {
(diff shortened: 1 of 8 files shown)


src/graphdb.js
```diff
@@ -195,7 +195,7 @@ function graphDBInferenceSchema (graphdbSchema, { addMutations = true, queryPref
                     if (direction.relationship === 'ONE-MANY') {
                         r += `\t${toCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${toCase}Input, options: Options, sort: [${toCase}Sort!]): [${toCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
                     }
-                    if (direction.relationship === 'MANY-ONE') {
+                    if (direction.relationship === 'MANY-ONE' || direction.relationship === 'ONE-ONE') {
```
kmcginnes, line 198: This is a good change. Would it make sense to make the many-many and one-many be consistent? Maybe even drop them in to a switch with a default case that logs. Something like this:

```typescript
switch (direction.relationship) {
    case 'MANY-MANY':
    case 'ONE-MANY':
        r += `\t${toCase.toLocaleLowerCase() + edgeCase}sOut(filter: ${toCase}Input, options: Options, sort: [${toCase}Sort!]): [${toCase}] @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
        break;
    case 'MANY-ONE':
    case 'ONE-ONE':
        r += `\t${toCase.toLocaleLowerCase() + edgeCase}Out: ${toCase} @relationship(edgeType:"${edge.label}", direction:OUT)\n`;
        break;
    default:
        loggerDebug(`Unknown relationship type: ${direction.relationship}`, {toConsole: true});
        break;
}
```

kmcginnes (approved):
LGTM

Could optionally go further to make the logic a bit more consistent.

PR #137, opened 2025-09-10
Handle non-nullable IDs for input types in resolver

andreachild:
Fixed issue where non-nullable id values were being passed as numbers in neptune query parameters by unwrapping non-nullable ID types for input types in resolver.

### DIFF
--- a/templates/JSResolverOCHTTPS.js
+++ b/templates/JSResolverOCHTTPS.js
@@ -74,7 +74,7 @@ export function resolveGraphDBQueryFromEvent(event) {
                 // retrieve an ID field which may not necessarily be named 'id'
                 const idField = Object.values(inputType.getFields()).find(
                     // if field is of non-null type, check the underlying type as well
-                    field => field?.type?.name === GraphQLID.name || field?.type?.ofType?.name === GraphQLID.name);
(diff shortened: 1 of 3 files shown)


templates/JSResolverOCHTTPS.js
```diff
@@ -64,11 +64,17 @@ export function resolveGraphDBQueryFromEvent(event) {
...
+                // if non-null type, unwrap to the underlying type
+                inputType = inputType.ofType;
+            }
             const astValue = astFromValue(value, inputType);
             if (inputType instanceof GraphQLInputObjectType) {
                 // retrieve an ID field which may not necessarily be named 'id'
-                const idField = Object.values(inputType.getFields()).find(field => field.type.name === GraphQLID.name);
+                const idField = Object.values(inputType.getFields()).find(
+                    // if field is of non-null type, check the underlying type as well
+                    field => field?.type?.name === GraphQLID.name || field?.type?.ofType?.name === GraphQLID.name);
```
kmcginnes, line 77:
```suggestion
                    field => field.type?.name === GraphQLID.name || field.type?.ofType?.name === GraphQLID.name);
```

I might be wrong, but I don't think `field` can ever be undefined or null here.

PR #157, opened 2026-04-06
Fix Lambda VPC config missing for Neptune DB with IAM auth

kirill-stepanishin:
### Description

   Neptune DB clusters always run inside a VPC, but the Lambda resolver was created without VPC configuration when IAM authentication was enabled. The Lambda could not reach the Neptune endpoint, making the `neptune-db + IAM` deployment combination non-functional.

   ### Root Cause

VPC configuration was conditionally applied based on the authentication method (`!NEPTUNE_IAM_AUTH`) rather than the database type (`NEPTUNE_TYPE === NEPTUNE_DB`). This incorrectly treated VPC and IAM as mutually exclusive, when in practice Neptune DB requires VPC connectivity regardless of how authentication is configured.

   ### Changes

   **Pipeline creation** (`src/pipelineResources.js`)
   - `createLambdaRole()`: split if/else into two independent blocks; neptune-db + IAM now gets all three policies (BasicExecution, NeptuneQuery, VPCAccess)
   - `createLambdaFunction()`: gate `VpcConfig` on `NEPTUNE_TYPE === NEPTUNE_DB` instead of `!NEPTUNE_IAM_AUTH`

   **CDK generation** (`src/CDKPipelineApp.js`, `templates/CDKTemplate.js`)
   - Restructure template from two separate `lambda.Function` calls to incremental props building with a single instantiation
   - Gate VPC config on `NEPTUNE_TYPE`, not auth method
   - Move `NEPTUNE_DBSubnetGroup` replacement inside the `neptuneClusterInfo` guard (previously produced string `'null'` for neptune-graph)

   **Cleanup** (`src/pipelineResources.js`)
   - Replace hardcoded policy detach blocks with dynamic loop over `LambdaExecutionPolicy*` keys

   **Error handling** (`src/pipelineResources.js`, `src/CDKPipelineApp.js`)
   - Always `exit(1)` when cluster info fetch fails for neptune-db, regardless of IAM flag (previously silently continued for IAM)
   - Improve error messages: "IAM authentication enabled/disabled" instead of "authentication is set to VPC/IAM"

   **Tests** (5 new files, 24 test cases)
   - Full 4-combination matrix (neptune-db/graph × IAM/no-IAM) for role creation, Lambda VpcConfig, policy cleanup, and CDK template output
   - Edge cases: partial cleanup failures, old resources.json format without Policy3, cluster info fetch failures

### DIFF
--- a/src/test/CDKPipelineApp.test.js
+++ b/src/test/CDKPipelineApp.test.js
@@ -1,0 +1,96 @@
+import { jest } from '@jest/globals';
+import { readFileSync } from 'fs';
+import { ExitCalled } from './helpers/pipelineMocks.js';
+
+const cdkTemplateContent = readFileSync('./templates/CDKTemplate.js', 'utf8');
+
+const mockGetNeptuneClusterDbInfoBy = jest.fn();
+const mockWriteFile = jest.fn();
+const mockReadFile = jest.fn();
+
+jest.unstable_mockModule('../pipelineResources.js', () => ({
+    getNeptuneClusterDbInfoBy: mockGetNeptuneClusterDbInfoBy,
+}));
+
+jest.unstable_mockModule('fs/promises', () => ({
+    readFile: mockReadFile,
+    writeFile: mockWriteFile,
+}));
+
+jest.unstable_mockModule('ora', () => ({
+    default: jest.fn(() => ({ start: jest.fn().mockReturnThis(), succeed: jest.fn(), fail: jest.fn(), warn: jest.fn() })),
+}));
+
+jest.unstable_mockModule('../zipPackage.js', () => ({
+    createLambdaDeploymentPackage: jest.fn(),
+}));
+
+jest.unstable_mockModule('../logger.js', () => ({
+    loggerDebug: jest.fn(), loggerError: jest.fn(), loggerInfo: jest.fn(), yellow: jest.fn(s => s),
+}));
+
+// CDKPipelineApp.js calls process.exit() directly (global), so spyOn is needed here
+const mockProcessExit = jest.spyOn(process, 'exit').mockImplementation((code) => { throw new ExitCalled(code); });
+
+let createAWSpipelineCDK;
+
+beforeAll(async () => {
+    ({ createAWSpipelineCDK } = await import('../CDKPipelineApp.js'));
+});
+
+beforeEach(() => {
+    jest.clearAllMocks();
+    mockReadFile.mockImplementation(async (path, enc) => {
+        if (path.includes('CDKTemplate')) return cdkTemplateContent;
+        return '';
+    });
+    mockWriteFile.mockResolvedValue(undefined);
+});
+
+const clusterInfo = {
+    host: 'db.cluster.us-east-1.neptune.amazonaws.com', port: '8182',
+    isIAMauth: true, version: '1.3.0.0', dbSubnetGroup: 'default-vpc-123',
+    dbSubnetIds: ['subnet-1', 'subnet-2'], vpcSecurityGroupId: 'sg-123',
+    iamPolicyResource: 'arn:aws:neptune-db:us-east-1:123:cluster-abc/*',
+};
+
+const baseParams = {
+    pipelineName: 'test', neptuneDBName: 'mydb', neptuneDBregion: 'us-east-1',
+    appSyncSchema: 'type Query { get: String }', lambdaFilesPath: '/tmp/lambda',
+    outputFile: '/tmp/output/cdk.js', __dirname: '.', quiet: true,
+    neptuneHost: 'db.cluster.us-east-1.neptune.amazonaws.com', neptunePort: '8182',
+    outputFolderPath: '/tmp/output', resolverFilePath: '/tmp/r.js', resolverSchemaFilePath: '/tmp/s.json',
+    schemaModel: { definitions: [{ kind: 'ObjectTypeDefinition', name: { value: 'Query' }, fields: [{ name: { value: 'get' } }] }] },
+};
+
+test('should exit on cluster info failure for neptune-db with IAM', async () => {
+    mockGetNeptuneClusterDbInfoBy.mockRejectedValue(new Error('cluster not found'));
+
+    try { await createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: true, neptuneType: 'neptune-db' }); } catch (e) { if (!(e instanceof ExitCalled)) throw e; }
+
+    expect(mockProcessExit).toHaveBeenCalledWith(1);
+});
+
+test('should exit on cluster info failure for neptune-db without IAM', async () => {
+    mockGetNeptuneClusterDbInfoBy.mockRejectedValue(new Error('cluster not found'));
+
+    try { await createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: false, neptuneType: 'neptune-db' }); } catch (e) { if (!(e instanceof ExitCalled)) throw e; }
+
+    expect(mockProcessExit).toHaveBeenCalledWith(1);
+});
+
+test('should generate CDK file for neptune-graph without calling cluster info', async () => {
+    await createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: true, neptuneType: 'neptune-graph' });
+
+    expect(mockGetNeptuneClusterDbInfoBy).not.toHaveBeenCalled();
+    expect(mockWriteFile).toHaveBeenCalled();
+    const writtenContent = mockWriteFile.mock.calls[0][1];
+    expect(writtenContent).toContain('NEPTUNE_IAM_AUTH = true');
+    expect(writtenContent).toContain('NEPTUNE_DBSubnetGroup = null');
+    expect(writtenContent).toContain('NEPTUNE_DBSubnetIds = null');
+    expect(writtenContent).toContain('NEPTUNE_VpcSecurityGroupId = null');
+});
+
+afterAll(() => {
+    mockProcessExit.mockRestore();
+});
--- a/src/test/CDKPipelineApp.test.js
+++ b/src/test/CDKPipelineApp.test.js
@@ -91,6 +91,30 @@ test('should generate CDK file for neptune-graph without calling cluster info', 
     expect(writtenContent).toContain('NEPTUNE_VpcSecurityGroupId = null');
 });
 
+test('should generate CDK file with VPC config for neptune-db with IAM', async () => {
+    mockGetNeptuneClusterDbInfoBy.mockResolvedValue(clusterInfo);
+    await createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: true, neptuneType: 'neptune-db' });
+    expect(mockWriteFile).toHaveBeenCalled();
+    const writtenContent = mockWriteFile.mock.calls[0][1];
+    expect(writtenContent).toContain("NEPTUNE_IAM_AUTH = true");
+    expect(writtenContent).toContain("NEPTUNE_DBSubnetGroup = 'default-vpc-123'");
+    expect(writtenContent).toContain("NEPTUNE_DBSubnetIds = 'subnet-1,subnet-2'");
+    expect(writtenContent).toContain("NEPTUNE_VpcSecurityGroupId = 'sg-123'");
+    expect(writtenContent).toContain("NEPTUNE_IAM_POLICY_RESOURCE = 'arn:aws:neptune-db:us-east-1:123:cluster-abc/*'");
+});
+
+test('should exit when neptune-db has IAM disabled but IAM flag was provided', async () => {
+    mockGetNeptuneClusterDbInfoBy.mockResolvedValue({ ...clusterInfo, isIAMauth: false });
+    try { await createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: true, neptuneType: 'neptune-db' }); } catch (e) { if (!(e instanceof ExitCalled)) throw e; }
+    expect(mockProcessExit).toHaveBeenCalledWith(1);
+});
+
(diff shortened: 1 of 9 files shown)


src/test/CDKTemplate.test.js
```diff
@@ -0,0 +1,104 @@
...
+
+function renderTemplate({ neptuneType, neptuneIAMAuth }) {
+    return templateSource
+        .replace("const NEPTUNE_TYPE = '';", `const NEPTUNE_TYPE = '${neptuneType}';`)
+        .replace("const NEPTUNE_IAM_AUTH = false;", `const NEPTUNE_IAM_AUTH = ${neptuneIAMAuth};`)
+        .replace("const NEPTUNE_DBSubnetGroup = null;", "const NEPTUNE_DBSubnetGroup = 'vpc-123';")
+        .replace("const NEPTUNE_DBSubnetIds = null;", "const NEPTUNE_DBSubnetIds = 'subnet-1,subnet-2';")
+        .replace("const NEPTUNE_VpcSecurityGroupId = null;", "const NEPTUNE_VpcSecurityGroupId = 'sg-123';")
+        .replace("const NEPTUNE_HOST = '';", "const NEPTUNE_HOST = 'db.cluster.us-east-1.neptune.amazonaws.com';")
+        .replace("const NEPTUNE_IAM_POLICY_RESOURCE = '*';", "const NEPTUNE_IAM_POLICY_RESOURCE = 'arn:aws:neptune-db:us-east-1:123:cluster-abc/*';");
```
kmcginnes: Since these types of string replacements tend to be brittle, I suggest throwing an exception, which would fail the test, if the string isn't found. That would force anyone changing that code/template to update the tests.

src/test/CDKTemplate.test.js
```diff
@@ -0,0 +1,104 @@
...
+    expect(managedPolicies).toContain('service-role/AWSLambdaVPCAccessExecutionRole');
+});
+
+test('neptune-graph with IAM: IAM policy, no VPC config', () => {
+    const { lambdaProps } = evaluateTemplate({ neptuneType: 'neptune-graph', neptuneIAMAuth: true });
+    expect(lambdaProps.vpc).toBeUndefined();
+    expect(lambdaProps.vpcSubnets).toBeUndefined();
+    expect(lambdaProps.securityGroups).toBeUndefined();
+    expect(lambdaProps.initialPolicy).toBeDefined();
+    expect(lambdaProps.initialPolicy[0].actions).toContain('neptune-graph:connect');
```
kmcginnes: Does this not need the same read query access as `neptune-db`?

```
neptune-graph:ReadDataViaQuery
```

src/test/CDKTemplate.test.js
```diff
@@ -0,0 +1,104 @@
...
+});
+
+test('neptune-graph without IAM: no VPC, no IAM policy', () => {
```
kmcginnes: Is this even possible? I thought Neptune Analytics required IAM always.

src/test/createPipeline.test.js
```diff
@@ -0,0 +1,164 @@
+import { jest } from '@jest/globals';
+import { mockIAMSend, mockLambdaSend, mockAppSyncSend, mockNeptuneSend, mockWriteFileSync, mockReadFileSync, mockExit, ExitCalled, registerPipelineMocks, mockNeptuneClusterResponse } from './helpers/pipelineMocks.js';
+
+registerPipelineMocks();
+
+const originalSetTimeout = globalThis.setTimeout;
```
kmcginnes: This could probably be a bit cleaner using Jest's built in fake timers.

https://jestjs.io/docs/timer-mocks

src/test/CDKPipelineApp.test.js
```diff
@@ -0,0 +1,96 @@
...
+    });
+    mockWriteFile.mockResolvedValue(undefined);
+});
+
+const clusterInfo = {
+    host: 'db.cluster.us-east-1.neptune.amazonaws.com', port: '8182',
+    isIAMauth: true, version: '1.3.0.0', dbSubnetGroup: 'default-vpc-123',
+    dbSubnetIds: ['subnet-1', 'subnet-2'], vpcSecurityGroupId: 'sg-123',
+    iamPolicyResource: 'arn:aws:neptune-db:us-east-1:123:cluster-abc/*',
+};
```
kmcginnes, line 55: This seems to be unused.

kmcginnes: Perhaps this was meant for a happy path test that seems to be missing? Kiro suggested:

```js
  test('should generate CDK file with VPC config for neptune-db with IAM', async () => {
      mockGetNeptuneClusterDbInfoBy.mockResolvedValue(clusterInfo);

      await createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: true, neptuneType: 'neptune-db' });

      expect(mockWriteFile).toHaveBeenCalled();
      const writtenContent = mockWriteFile.mock.calls[0][1];
      expect(writtenContent).toContain("NEPTUNE_IAM_AUTH = true");
      expect(writtenContent).toContain("NEPTUNE_DBSubnetGroup = 'default-vpc-123'");
      expect(writtenContent).toContain("NEPTUNE_DBSubnetIds = 'subnet-1,subnet-2'");
      expect(writtenContent).toContain("NEPTUNE_VpcSecurityGroupId = 'sg-123'");
  });
```

src/test/CDKPipelineApp.test.js
```diff
@@ -0,0 +1,96 @@
...
+    expect(mockProcessExit).toHaveBeenCalledWith(1);
+});
+
+test('should exit on cluster info failure for neptune-db without IAM', async () => {
+    mockGetNeptuneClusterDbInfoBy.mockRejectedValue(new Error('cluster not found'));
+
+    try { await createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: false, neptuneType: 'neptune-db' }); } catch (e) { if (!(e instanceof ExitCalled)) throw e; }
+
+    expect(mockProcessExit).toHaveBeenCalledWith(1);
+});
```
kmcginnes, line 80: Missing two of the test scenarios that you have in `createPipeline.test.js`.

```js
  test('should exit when neptune-db has IAM disabled but IAM flag was provided', async () => {
      mockGetNeptuneClusterDbInfoBy.mockResolvedValue({ ...clusterInfo, isIAMauth: false });

      await expect(createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: true, neptuneType: 'neptune-db' }))
          .rejects.toThrow(ExitCalled);
      expect(mockProcessExit).toHaveBeenCalledWith(1);
  });

  test('should exit when neptune-db has IAM enabled but IAM flag was not provided', async () => {
      mockGetNeptuneClusterDbInfoBy.mockResolvedValue({ ...clusterInfo, isIAMauth: true });

      await expect(createAWSpipelineCDK({ ...baseParams, isNeptuneIAMAuth: false, neptuneType: 'neptune-db' }))
          .rejects.toThrow(ExitCalled);
      expect(mockProcessExit).toHaveBeenCalledWith(1);
  });
```

src/test/createPipeline.test.js
```diff
@@ -0,0 +1,164 @@
...
+        expect(policyDoc.Statement[0].Action).toContain('neptune-graph:connect');
+    });
+
+    test('should attach only basic execution role for neptune-graph without IAM', async () => {
+        await createUpdateAWSpipeline({ ...baseParams, isNeptuneIAMAuth: false, neptuneType: 'neptune-graph' });
+
+        const attachCalls = mockIAMSend.mock.calls.filter(([c]) => c._type === 'AttachRolePolicy').map(([c]) => c.input.PolicyArn);
+        expect(attachCalls).toContain('arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole');
+        expect(attachCalls).toHaveLength(2); // basic + invoke policy
+        expect(attachCalls).not.toContain('arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole');
```
kmcginnes:
```suggestion
        expect(attachCalls).toContain('arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole');
        expect(attachCalls).not.toContain('arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole');
        expect(attachCalls).toHaveLength(2); // basic + invoke policy
```

I would at the very least re-order these so if this test was to fail it would provide a decent message, rather than an unhelpful "expected 2, but got 3". You could also just drop the length check, or check the full list of policies is exactly some specific list.

src/test/CDKTemplate.test.js
```diff
@@ -0,0 +1,104 @@
...
+        .replace("const NEPTUNE_IAM_POLICY_RESOURCE = '*';", "const NEPTUNE_IAM_POLICY_RESOURCE = 'arn:aws:neptune-db:us-east-1:123:cluster-abc/*';");
+}
+
+function evaluateTemplate({ neptuneType, neptuneIAMAuth }) {
+    const code = renderTemplate({ neptuneType, neptuneIAMAuth });
+    // Extract from "const lambdaProps = {" up to "echoLambda = new lambda.Function"
+    const match = code.match(/const lambdaProps = \{[\s\S]*?(?=\n\s*echoLambda = new lambda\.Function)/);
+    if (!match) throw new Error('Could not extract lambdaProps block from CDKTemplate.js — check that "const lambdaProps = {" and "echoLambda = new lambda.Function" markers still exist in the template');
+
+    const managedPolicies = [];
```
kmcginnes: Seems unused

src/test/removePipeline.test.js
```diff
@@ -0,0 +1,169 @@
...
+import { mockIAMSend, mockLambdaSend, mockAppSyncSend, registerPipelineMocks } from './helpers/pipelineMocks.js';
+
+registerPipelineMocks();
+
+let removeAWSpipelineResources;
+
+describe('removeAWSpipelineResources', () => {
+
+beforeEach(async () => {
+    ({ removeAWSpipelineResources } = await import('../pipelineResources.js?t=' + Date.now()));
```
kmcginnes, line 12:
```suggestion
    // Re-import with unique query string to reset module-level mutable state between tests
    ({ removeAWSpipelineResources } = await import('../pipelineResources.js?t=' + Date.now()));
```

src/test/createPipeline.test.js
```diff
@@ -0,0 +1,164 @@
...
+registerPipelineMocks();
+
+const originalSetTimeout = globalThis.setTimeout;
+
+let createUpdateAWSpipeline;
+
+beforeEach(async () => {
+    globalThis.setTimeout = (fn) => originalSetTimeout(fn, 0);
+
+    ({ createUpdateAWSpipeline } = await import('../pipelineResources.js?t=' + Date.now()));
```
kmcginnes, line 13:
```suggestion
    // Re-import with unique query string to reset module-level mutable state between tests
    ({ createUpdateAWSpipeline } = await import('../pipelineResources.js?t=' + Date.now()));
```

kmcginnes (commented):
Mostly suggested cleanup and minor improvements.

This change does not account for Neptune DB public endpoints, which are now a thing. They require IAM to be enabled, but do not require a VPC. I would assume this is out of scope for this change, but worth noting.

src/test/CDKTemplate.test.js
kmcginnes: I recently had to do this.

https://github.com/aws/graph-explorer/pull/1640/changes#diff-0a05b1ab392db457fd15d2ff2f21897444be01c2272d8e28dcdf282d6a4d2575R20

kirill-stepanishin: Done. added a `safeReplace` helper that throws if the placeholder is missing. Thanks for the reference.

src/test/CDKTemplate.test.js
kirill-stepanishin: Fixed, both IAM tests now use `toEqual` with all 4 actions.

src/test/CDKTemplate.test.js
kirill-stepanishin: True, this test is not very useful. Removed it and equivalents in other test files. Thanks!

src/test/CDKPipelineApp.test.js
kirill-stepanishin: Thank you; added the happy-path test that uses it, including IAM policy resource assertion.

src/test/CDKPipelineApp.test.js
kirill-stepanishin: Added both. Thanks!

src/test/CDKTemplate.test.js
kirill-stepanishin: Removed.

src/test/createPipeline.test.js
kirill-stepanishin: Good idea, applied your suggestion. Thank you

src/test/removePipeline.test.js
kirill-stepanishin: Applied this wording to re-imports across tests. Thank you

src/test/createPipeline.test.js
kirill-stepanishin: Refactored using `jest.useFakeTimers()`. Thanks!

kirill-stepanishin:
> This change does not account for Neptune DB public endpoints, which are now a thing. They require IAM to be enabled, but do not require a VPC. I would assume this is out of scope for this change, but worth noting.

This change addresses exactly this scenario. Neptune DB public endpoints are still gated by security groups, which typically only allow traffic from within the VPC or specific IPs. A Lambda outside the VPC can't connect unless the security group is opened to all IPs, which we shouldn't encourage. Placing the Lambda in the VPC works for both public and private endpoints.

kmcginnes (approved):

PR #160, opened 2026-04-07
Update Axios and AWS SDK dependencies

kirill-stepanishin:
### Description
Update Axios & AWS SDK dependencies across root and test directories.


test/package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "graphql-tag": "2.12.6",
     "adm-zip": "0.5.16",
-    "@aws-sdk/client-appsync": "^3.445.0",
-    "axios": "^1.12.0"
+    "@aws-sdk/client-appsync": "^3.1025.0",
+    "axios": "^1.14.0"
```
kmcginnes: It looks like you are in the rat race with me. A new Axios version was released 2 days ago with more fixes.

https://github.com/axios/axios/releases/tag/v1.15.0

kmcginnes (approved):
It looks like there is another round of important updates to do for Axios, but that can either be part of this PR or a follow up PR. Your call.

test/package.json
kirill-stepanishin: Thank you, Bumped AWS SDK and Axios to latest versions

PR #162, opened 2026-04-16
Fix remove field prefix matching bug and add tests

kirill-stepanishin:
## Problem

`removeChanges()` used `line.startsWith(change.field)` to match fields for removal. This caused prefix collisions — removing `desc` would also remove `desc2`, `description`, etc.

## Fix

Require the field name to be followed by a terminating character (`:`, `(`, or ` `), preventing partial matches.

## Tests

Added unit tests for `changeGraphQLSchema` covering:
- Core add/remove functionality
- Prefix collision fix verification
- Edge cases (empty field, no-op changes, directive preservation, type scoping)

### DIFF
--- a/src/test/changes.test.js
+++ b/src/test/changes.test.js
@@ -1,0 +1,136 @@
+import { changeGraphQLSchema } from '../changes.js';
+
+const baseSchema = `
+type Airport {
+    desc: String
+    desc2: String
+    code: String
+    getAirport(code: String): Airport
+}
+
+type Query {
+    getAirport(code: String): Airport
+    listAirports: [Airport]
+}
+`;
+
+// Mimics changeGraphQLSchema() internal formatting: trim each line and append a trailing newline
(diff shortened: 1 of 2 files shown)


src/test/changes.test.js
```diff
@@ -0,0 +1,136 @@
...
+// Mimics changeGraphQLSchema() internal formatting: trim each line and append a trailing newline
+const normalize = (s) => s.split('\n').map(l => l.trim()).join('\n') + '\n';
+
+describe('changeGraphQLSchema', () => {
+    // Core functionality
+
+    test('add action injects field before closing brace of matching type', () => {
+        const changes = JSON.stringify([{ type: 'Airport', action: 'add', value: 'city: String' }]);
+        const result = changeGraphQLSchema(baseSchema, changes);
+        expect(result).toContain('city: String');
```
kmcginnes: This doesn't verify that the change was injected before the closing brace. It could've been appended at the end of the string.

kmcginnes (approved):
Minor nits, nothing major.

PR #163, opened 2026-04-20
Add custom type definitions support in schema changes file

### LINKED ISSUE #156: Support adding new GraphQL type definitions via `--input-schema-changes-file`
## Description                                                                                   
                                         
  The `--input-schema-changes-file` option currently supports adding/removing fields on existing   
  GraphQL types, but does not support creating entirely new type definitions. This is a limitation
  when defining custom queries (via `@graphQuery`) that return a type not already present in the   
  schema derived from the graph database.                                                          

  ## Current Behavior

  The `changeGraphQLSchema` function in `src/changes.js` iterates through schema lines, matches the
   `"type"` field in the JSON directive to existing `type <Name> {` declarations, and injects new
  fields before the closing `}`. If the `"type"` value doesn't match any existing type in the
  schema, the entry is silently ignored.

  For example, the following entry has no effect because there is no `type
  PersonWithAccountsResult` in the generated schema:

  ```json
  {
    "type": "PersonWithAccountsResult",
    "field": "PersonWithAccountsResult",
    "action": "add",
    "value": "type PersonWithAccountsResult { id: ID! first_name: String last_name: String }"
  }
 ```
  **Use Case**

  When adding a custom query that returns a composite/projected result (not a direct node type from
   the graph), a new return type is needed:

```json 
 [
    {
      "type": "Query",
      "field": "getPersonWithAccounts",
      "action": "add",
      "value": "getPersonWithAccounts(name: String!): PersonWithAccountsResult @graphQuery(statement: \"...\")"
    }
  ]
```

  This adds the query field to Query successfully, but there is no way to define
  PersonWithAccountsResult through the same file — requiring a manual post-processing step to
  append the type to the generated schema.

  Suggested Enhancement

  Support a new action (e.g., "action": "addType") or a convention where if the "type" value
  doesn't match an existing type, a new top-level type definition is appended to the schema. For
  example:

```json
  {
    "type": "PersonWithAccountsResult",
    "action": "addType",
    "value": "type PersonWithAccountsResult { id: ID! first_name: String last_name: String }"
  }
```
  This would allow --input-schema-changes-file to be fully self-contained for custom query
  scenarios without requiring external schema manipulation.

kirill-stepanishin:
Adds a new `addType` action to `--input-schema-changes-file` for defining custom GraphQL types during schema generation.

Previously, custom `@graphQuery` directives returning composite results required manual schema editing after generation, because the changes file could only add/remove fields on existing types and not define new ones.

Now a changes file can define the return type and the query together:

```json
[
  {
    "action": "addType",
    "value": "type AirportStats {\n  count: Int\n  name: String\n}"
  },
  {
    "type": "Query",
    "field": "getAirportStats",
    "action": "add",
    "value": "getAirportStats: AirportStats @graphQuery(statement: \"MATCH (a:airport) RETURN {count: count(a), name: 'airports'} AS getAirportStats\")"
  }
]
```

The first entry creates `AirportStats`, the second adds a query returning it. A return type validator runs before parsing and suggests `addType` when Query/Mutation fields reference undefined types.

Other changes:
- Extracted `AWS_APPSYNC_SCALARS` to shared constant in `src/util.js`
- Input validation for malformed JSON and non-array payloads in changes file
- Offline integration test (Case11)
- README and CLI help updated

Closes #156


src/test/changes.test.js
```diff
@@ -135,3 +148,328 @@ describe('changeGraphQLSchema', () => {
...
+    });
+
+    test('addType with empty string value does not corrupt the schema', () => {
+        const changes = JSON.stringify([{ action: 'addType', value: '' }]);
+        const result = changeGraphQLSchema(baseSchema, changes);
+        expect(result).toBe(normalize(baseSchema));
+    });
+});
+
+describe('validateReturnTypes', () => {
```
kmcginnes: My main concern with these tests is that they only cover the passing validation side of many scenarios. We need to include the failing validation side to ensure the logic is performing any kind of check.

So, I asked Kiro to review through these tests and summarize the missing tests & scenarios, and offer some organization suggestions (organization is just a nit).

---

## `validateReturnTypes` test review

### Missing negative counterparts

Several "passes" tests don't have a corresponding "throws" version. The wrapper-unwrapping logic (`!`, `[]`) is only tested in the positive direction — if `extractReturnType` had a bug stripping these, only negative tests would catch it.

| Passes test | Has negative? |
|---|---|
| Non-null type (`Foo!`) | ❌ No test for `Missing!` throwing |
| Non-null list (`[Foo!]`) | ❌ No test for `[Missing!]` throwing |
| Nested non-null list (`[Foo!]!`) | ❌ Not tested at all |
| Enum return type | ❌ |
| Interface return type | ❌ |
| Union return type | ❌ |
| Input return type | ❌ |
| User-defined scalar | ❌ |
| `extend type Foo` defines Foo | ❌ |
| Directive on Query type line | ❌ |
| Indented schema | ❌ |

### Missing scenarios

- **Mutation with list/non-null return types** — all list/non-null tests use `Query` only
- **Both Query and Mutation in the same schema** with errors in each
- **`extend type Mutation`** — only `extend type Query` is tested
- **Nested non-null list** like `[Foo!]!`
- **Malformed field with no return type** (e.g. `getFoo`) inside Query — `extractReturnType` returns null for this, but it's not tested through `validateReturnTypes`

### Organization

The 38 tests are in a flat list mixing unrelated concerns. Suggested grouping:

1. **Input guards** — null, undefined, empty schema
2. **Type collection** — type, enum, input, scalar, interface, union, extend, comments, descriptions
3. **Wrapper unwrapping** — `[T]`, `T!`, `[T!]`, `[T!]!` (use `test.each`)
4. **Query/Mutation block detection** — directives on type line, single-line declarations, extend, both blocks
5. **Field parsing** — directives, default values, inline comments, multi-line descriptions
6. **Scalars** — built-in, AWS, user-defined (use `test.each`)
7. **Error reporting** — single missing type, multiple missing types, error message format

### Condensation opportunities

These groups of near-identical tests could each become a single `test.each`:

- The 4 non-string `addType` value tests (null, number, object, undefined)
- The 3 wrapper unwrapping tests (`[Foo]`, `Foo!`, `[Foo!]`)
- The 3 scalar tests (built-in, AWS, user-defined)
- Several "passes with a defined type" tests that only differ in schema syntax

src/changes.js
```diff
@@ -45,7 +49,15 @@ function removeChanges(changesDirectives, currentType, line) {
 
 
 function changeGraphQLSchema(schema, changes) {
-    const changesDirectives = JSON.parse(changes);
+    let changesDirectives;
+    try {
+        changesDirectives = JSON.parse(changes);
+    } catch (err) {
+        throw new Error('Invalid JSON in --input-schema-changes-file: ' + err.message);
```
kmcginnes:
```suggestion
        throw new Error('Invalid JSON in --input-schema-changes-file: ' + err.message, { cause: err });
```

It's good practice to include the original error in the `cause` field when rethrowing.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

src/changes.js
```diff
@@ -72,8 +84,134 @@ function changeGraphQLSchema(schema, changes) {
         }
     }
 
+    for (const change of changesDirectives) {
+        if (change.action == 'addType') {
+            if (typeof change.value === 'string' && change.value.trim()) {
+                r += change.value + '\n';
+            } else {
+                loggerInfo('Skipping addType entry with missing or non-string value');
```
kmcginnes: Should this be a warning, or is that too strong?

src/help.js
```diff
@@ -88,12 +88,12 @@ Input a property graph database schema
...
  - the user can remove or add to the GraphQL schema generated by the utility.
  - Changes file format: [
-     { "type": "graphQLTypeName", 
-       "field": "graphQLFieldName",       
-       "action": "remove|add",
-       "value": "value"
-    }   
+     { "type": "graphQLTypeName", "field": "graphQLFieldName", "action": "add",    "value": "fieldDefinition" },
+     { "type": "graphQLTypeName", "field": "graphQLFieldName", "action": "remove" },
+     { "action": "addType", "value": "type NewType { field: String }" }
```
kmcginnes:
```suggestion
     { "action": "add", "type": "graphQLTypeName", "field": "graphQLFieldName",    "value": "fieldDefinition" },
     { "action": "remove", "type": "graphQLTypeName", "field": "graphQLFieldName" },
     { "action": "addType", "value": "type NewType { field: String }" }
```

The order of the fields should at least be consistent. I suggest, putting them in this order with action first.

README.md
```diff
@@ -400,12 +400,29 @@ applied at once.
...
-    "action": "remove|add",
-    "value": "value"
+    "action": "add",
+    "value": "fieldDefinition"
+  },
+  {
+    "type": "graphQLTypeName",
+    "field": "graphQLFieldName",
+    "action": "remove"
+  },
```
kmcginnes, line 410:
```suggestion
  {
    "action": "remove",
    "type": "graphQLTypeName",
    "field": "graphQLFieldName"
  },
```

kmcginnes (commented):
This seems mostly fine, although there are some missing test cases that need to be added.

src/changes.js
```diff
@@ -72,8 +84,134 @@ function changeGraphQLSchema(schema, changes) {
...
+/**
+ * Checks that Query/Mutation return types are defined. Line-based, so it
+ * misses some edge cases (multi-line directive args, multi-field single-line blocks).
+ */
+function validateReturnTypes(schema) {
+    if (!schema) return;
+
+    const knownTypes = new Set([
+        ...specifiedScalarTypes.map(t => t.name),
+        ...AWS_APPSYNC_SCALARS
```
Cole-Greer, line 110: Do we need to exclude `AWS_APPSYNC_SCALARS` from known types if the target is Apollo? Or do we automatically add `AWS_APPSYNC_SCALARS` types in all cases?

src/test/changes.test.js
kirill-stepanishin: Done. Added negative counterparts, reorganized into describe blocks, consolidated with `test.each`. Also migrated `validateReturnTypes` to AST

src/changes.js
kirill-stepanishin: Done, thanks

src/changes.js
kirill-stepanishin: There is no `loggerWarn` in the codebase, changed to `loggerInfo({toConsole: true})` so the user sees it.

src/help.js
kirill-stepanishin: Done

README.md
kirill-stepanishin: Done, thanks

src/changes.js
kirill-stepanishin: `validateReturnTypes` runs before we know the target (AppSync vs Apollo). Apollo-specific scalar validation can be a separate enhancement. I think leaving it as-is is reasonable for now.

kmcginnes (approved):

PR #169, opened 2026-05-05
Bump axios to 1.16.0 and regenerate third-party licenses

kirill-stepanishin:
Updates `axios` across root, Lambda4AppSyncHTTP, and ApolloServer packages. Regenerates `THIRD_PARTY_LICENSES.txt` by running `@electrovir/oss-attribution-generator` separately in root + 3 Lambda templates, then merging and deduplicating the results into a single file.

Closes #168


kmcginnes (approved):
Approving version bump. I'm assuming the third party license file is correct since it is auto-generated from the dependencies.

PR #175, opened 2026-06-05
Use parameterized queries for @graphQuery/@cypher argument substitution

theneelshah:
Arguments are passed as Neptune query parameters instead of being inlined into the query string.


kmcginnes (commented):
I think this looks fine as a change. My only questions are:

* **Did we go far enough?** Are there more places in the code that inline arguments that should instead be parameterized?
* **How are collisions handled?** From my scan of the code it seems like we have last one wins, which is fine. How likely are we to have collisions for parameter names?

templates/JSResolverOCHTTPS.js
```diff
@@ -512,7 +512,9 @@ function createQueryFunctionMatchStatement(obj, matchStatements, querySchemaInfo
     if (querySchemaInfo.graphQuery != null) {
         var gq = querySchemaInfo.graphQuery.replaceAll('this', querySchemaInfo.pathName);
         obj.definitions[0].selectionSet.selections[0].arguments.forEach(arg => {
-            gq = gq.replace('$' + arg.name.value, arg.value.value);
+            const paramName = querySchemaInfo.pathName + '_' + arg.name.value;
+            gq = gq.replace(`'$${arg.name.value}'`, `$${paramName}`);
```
kirill-stepanishin: This replace pattern requires single quotes around the placeholder, which won't match unquoted params. An example is `{dist:$dist}` in `addRoute` (`test/TestCases/airports.source.schema.graphql:256`).

kirill-stepanishin: This made me realize `.replace()` here only substitutes the first occurrence. If a graphQuery reuses the same `$param` twice, the second one is left unreplaced. Since we're already touching this code, might be a good time to swap to `.replaceAll()`

templates/JSResolverOCHTTPS.js
```diff
@@ -1255,7 +1257,9 @@ function resolveGraphDBqueryForGraphQLMutation (queryAst, querySchemaInfo) {
             ocQuery = ocQuery.replace('$input', formattedFields);
         } else {
             queryAst.definitions[0].selectionSet.selections[0].arguments.forEach(arg => {
-                ocQuery = ocQuery.replace('$' + arg.name.value, arg.value.value);
+                const paramName = querySchemaInfo.pathName + '_' + arg.name.value;
+                ocQuery = ocQuery.replace(`'$${arg.name.value}'`, `$${paramName}`);
```
kirill-stepanishin:
```suggestion
                ocQuery = ocQuery.replaceAll(`'$${arg.name.value}'`, `$${paramName}`);
```

kirill-stepanishin (commented):
Good change overall, I left a couple of suggestions. Additionally, we should consider applying the same parameterization to the Gremlin path (still inlines values directly on line 1414), and ideally adding test coverage for the modified mutation code path since it's currently untested.

kirill-stepanishin (approved):
LGTM, thanks!

kmcginnes (approved):

